PR #4, opened 2021-10-14
fixes for big number vis

gau94rav:


visualizations/big_number.html
```diff
@@ -15,7 +15,7 @@
                         :style='{color: arrowColor}'
                     ></i>
                 </span>
-                <span v-if='value !== previous' class="opacity-70">
+                <span v-if='valueRaw !== prevRaw' class="opacity-70">
```
JARosen, line 18: This is the only part I don't get.  If the user wants to show the change, and the values are the same, it should show a -.  But the whole span will be hidden, right?  I think if there is a change column, then we should always show the change, even if it's -.

PR #49, opened 2022-03-10
Tag cleanup

dbroudy:
I think some tags got mis-nested and the stricter html parser is flagging a couple things:

Screen Shot 2022-03-10 at 1 29 00 PM


gau94rav (approved):
LGTM, thank you @dbroudy for fixing these.

PR #50, opened 2022-03-21
 Updated events to kebab-case instead of camelCase. 

msorber:
Also renamed "page"  data variable in TPager to "internalPage" since TopCoat is adding a "page" variable as well.


gau94rav:
These are the components that are present in expandable-modules.
1. Text Input (Pushed to public)
2. Select Input
3. Multi Select Input
4. Grids
5. Tabs

PR #71, opened 2022-06-13
Table styling

msorber:
Need to support table styling that matches the Issues report design. As part of that some columns on the front end combine data from 2 columns on the back end, so added a new "exludeFromColumns" prop that allows the user to keep a column of data in the data set, but not have a column created for it. The front end column that combines the 2+ back end columns can still access the data through the row object passed to the column.

Also added props for headerClasses and cellClasses.

And made the spinner show up on top of the rows while the table is fetching the next page of data instead of replacing the rows, that way the height of the table doesn't jump around an insane amount.

### DIFF
--- a/library/components/SnykPager.vue
+++ b/library/components/SnykPager.vue
@@ -1,0 +1,176 @@
+<template>
+  <div>
+    <div
+      class="pagination-footer flex justify-center relative"
+      v-if="numberOfItems"
+    >
+      <ul class="pagination-nav pagination-selector inline-flex pt-[16px]">
+        <li class="pagination-nav-item">
+          <button
+            class="
+              h-8
+              minWidth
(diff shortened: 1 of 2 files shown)


library/components/SnykPager.vue
```diff
@@ -0,0 +1,176 @@
...
+    endIndex() {
+      let endIndex = this.startIndex + this.itemsPerPage;
+      if (endIndex > this.numberOfItems) endIndex = this.numberOfItems;
+      return endIndex;
+    },
+    isValidPage() {
+      return this.validatePage(this.internalPage);
+    },
+  },
+  watch: {
```
gau94rav, line 134: @msorber watchers usually are problematic when it comes to performance but if it's absolutely required, go for it.

gau94rav (approved):
LGTM

PR #75, opened 2022-06-23
feat: Do not show the header if report is running on an iframe

egcastro:
To be able to remove the top header nav when we are running the report in an iframe, I'm changing the default value to false when in an iframe and true otherwise.
We need to remove this prop from here since the default will be true from now on: https://github.com/snyk/snyk-insights-topcoat/blob/main/pages/issues_detail.html#L3

### DIFF
--- a/library/templates/Report/Report.vue
+++ b/library/templates/Report/Report.vue
@@ -32,10 +32,11 @@
       popup:false
     }),
     props: {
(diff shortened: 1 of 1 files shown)


dbroudy (approved):
LGTM

gau94rav (approved):
Looks good.

PR #78, opened 2022-06-30
Table cleanup

msorber:
Goals:
Add component for red x/ green check depending on prop of which to show (check patch UI)

Add component for open/fixed depending on prop (check patch UI)

Add component for severity level

date localization

Increase size of spinner

Increase space between Table title and the table

add “word-break: break-word” styling to tables cells

fix snyk pager “page of total pages” overlapping the paging buttons making them hard to click

Fix SnykPager not working correctly with <5 pages

add minimum height/width to table


library/components/CheckOrX.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div class="centerCheckOrX">
+    <span v-if="value" class="text-[#2D9283] mr-[6px]">
+      <i class="fas fa-check-circle"> </i>
```
gau94rav, line 4: Same as above, conditional classes.

library/components/SnykPager.vue
```diff
@@ -93,26 +96,36 @@ export default {
       pageNumbersToShow = [];
       let pageToAdd = this.internalPage;
       let numberOfPagesAdded = 0;
-      while (pageToAdd <= this.lastPage && numberOfPagesAdded < 4) {
-        pageNumbersToShow.push(pageToAdd);
-        pageToAdd++;
-        numberOfPagesAdded++;
+
+      // if there are only up to 5 pages, only show the valid pages, do
```
gau94rav, line 100: Yeah I remember struggling with this once 😅

visualizations/frontend/other/TTable.vue
```diff
@@ -170,7 +170,7 @@
     </div>
 
     <div style="margin: 0px auto">
-      <SnykPager
+      <TestSnykPager
```
gau94rav, line 173: Temporary file?

gau94rav (approved):
LGTM, though one small point, any component that is exclusive to table alone, I think should be moved to a separate folder (`/Table`).

library/components/CheckOrX.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div class="centerCheckOrX">
+    <span v-if="value" class="text-[#2D9283] mr-[6px]">
```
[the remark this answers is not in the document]
msorber: That is simpler! I normally avoid ternaries when the output can't be assigned to a descriptive variable, but in this case it is simple enough to be pretty obvious with the shorter approach. :D

library/components/SnykPager.vue
msorber: Yeah... It was a teensy bit irritating to realize I'd missed that case earlier 😅 Glad I'm not alone!

visualizations/frontend/other/TTable.vue
msorber: ....yup. Great catch thank you!

PR #88, opened 2022-07-11
fixed general purpose csv download

msorber:
As part of the fix for the back end running out of memory because the table and csv download were both requesting the entire data set without a limit I moved the TscvExport component from the visualizations folder to the library folder so the data is not automatically fetched. However that broke the csv download, so this is part of the fix.


gau94rav (approved):
LGTM, thanks for fixing this @msorber

PR #89, opened 2022-07-13
Create TUrl.vue

msorber:
The snyk-insights table needs a way to open a url to the issue page which will mean combining a base url with specific data from the table row
The executive summary page will need to open another page with almost all the current filters, except for 1 specific filter

### DIFF
--- a/library/components/TUrl.vue
+++ b/library/components/TUrl.vue
@@ -1,6 +1,6 @@
 <template>
   <a :href="fullUrl" class="tUrl">
-    <slot ></slot>
+    <slot></slot>
   </a>
 </template>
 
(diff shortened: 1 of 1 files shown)


JARosen (approved):
Nice and simple.  Great job!

sareg0:
Interesting. I think I would have gone with a 'Link'. Would the name 'Link' have conflicted at all with anything else? I think that `URL` has a different meaning in the web space, to 'Link'

Here is the ProductCL Base anchor component for inspiration. https://github.com/PatchUI/ProductCL/blob/main/src/components/BaseAnchor/BaseAnchor.vue


Is there any way to document this component right now? e.g. in Archbee?

PR #90, opened 2022-07-14
Add :openInNewTab" prop to the TUrl component

msorber:
The insights table should open info on specific issues in a new tab, added a prop to support that to the TUrl component.


dbroudy (approved):
Got the `rel` security stuff too, nice job!

PR #94, opened 2022-07-18
feat: add horizontal scroll on table

sareg0:
We have been intending to add the horizontal scroll on the table for a while. This is the most basic implementation of that, so that tables with lots of columns aren't squished, or overflowing their container.


msorber (approved):
LGTM and does work, but we may want to add a min-width to the table as well because the columns will shrink a lot before the scrolling kicks in.

PR #103, opened 2022-07-29
feat: Adapt code to new architecture

egcastro:
Related PR: https://github.com/snyk/topcoat-core/pull/469

### DIFF
--- a/library/components/TDropdown.vue
+++ b/library/components/TDropdown.vue
@@ -6,21 +6,21 @@
 		ref="dropdownFilter"
 	>
         <div
-            @click="setPopup(!popup)" :class="activeClass"
+            @click="openPopup" :class="activeClass"
             class="rounded cursor-pointer"
         >
             <slot name="handle"></slot>
         </div>
 		<slot name="outside"></slot>
 		<div
-			v-show="popup"
+			v-show="isPopupOpen"
 			ref="popup"
 			class="absolute mt-1 bg-white rounded-lg shadow-md base-dropdown-menu w-max min-w-"
 			:class="alignClass"
 			:style="{ zIndex }"
             :key="componentKey"
 		>
-            <slot :popup="setPopup"></slot>
+            <slot></slot>
 		</div>
 	</div>
 </template>
@@ -28,10 +28,6 @@
 <script>
 	export default {
         props: {
-            disableHandleClass: {
-                type: Boolean,
-                default: false,
-            },
             isActive: {
                 type: Boolean,
                 default: false,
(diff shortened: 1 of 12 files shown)


library/components/TDropdown.vue
```diff
@@ -43,73 +39,88 @@
...
 			},
-			setPopup(toggle) {
-				this.popup = toggle;
-				this.internal_active = toggle;
-
-				if (toggle) {
-					document.body.addEventListener("click", this.handleOutsideClick);
-				} else {
-					document.body.removeEventListener("click", this.handleOutsideClick);
+			openPopup() {
```
sareg0, line 102:
```suggestion
			togglePopup() {
```

Non-blocking for merging: only a suggestion to better capture that the function handles closing the popup as well as opening

visualizations/frontend/filters/v2/TMultiSelector.vue
```diff
@@ -173,6 +170,9 @@
             updateUrlParam() {
                 this.setFilterValue("selected_items", this.checked.join('|'), true);
             },
+            onDropdownOpen() {
+                this.fetchLayerData();
+            }
```
sareg0, line 175: 💅 💪 ✨

library/components/TDropdown.vue
```diff
@@ -43,73 +39,88 @@
...
-				})
+				});
 			},
 		},
+		watch: {
+			isExpanded: function(newVal, oldVal) {
+				if (newVal) {
+					this.isPopupOpen = true;
+					this.$emit('open');
+				}
```
sareg0, line 121: Do we need an else to emit 'close'?

egcastro: Not really. This is quite complicated to understand by the way we are wiring components with the render function (Vue's render function) and scoped slots. The trick here is that we can re-use the same code path that this component's parent will follow if the `@open` callback is been called when a props like `isExpanded` is change. But this is due to the fact that we are rendering components with this prop `is-expanded` set to true already. Meaning that there was never a `click` and no way to trigger the `fetchLayerData` callback programatically. This is a way of simulating a `click` programatically based on a prop set/change 😄

egcastro: But the `close` is only trigger but the clicking outside. There is no way to pass a prop for closing the the dropdown.
This is mostly a bigger topic that we should discuss at some point, components' API and behaviours in general.

sareg0: ah, ok. I thought there was probably a good reason, but felt compelled to ask.

gau94rav (approved):
LGTM.

PR #106, opened 2022-08-04
feat: Adding support for default-value in TSelect and TMultiSelector

egcastro:
We can use `default-value` as a `string` in `<t-multi-selector>` and `<t-select>` to set the value and its corresponding filter state in the URL.

Example:
```html
<t-multi-selector is-expanded default-value="False" t-layer="filter_ignored" t-filter:selected_items="ignored" item-label="Currently Ignored" />
```

```html
<t-select default-value="Severity" t-layer="filter_layer_type" t-filter:dropdown="issue_by" label="Issues By" item-label="Issues_by" />
```

### DIFF
--- a/visualizations/frontend/filters/v2/TMultiSelector.vue
+++ b/visualizations/frontend/filters/v2/TMultiSelector.vue
@@ -150,13 +150,16 @@
 				}
 				return attrs;
 			}
-		},
+        },
 		methods: {
             onVisualizationInit() {
 				const initial_value = this.getFilterValue("selected_items");
 
                 if (initial_value) {
                     this.checked = initial_value.split('|');
+                } else if (this.defaultValue) {
+                    this.checked = this.defaultValue.split('|');
+                    this.setFilterValue("selected_items", this.defaultValue);
                 }
             },
 			selectUnselect() {
@@ -168,7 +171,7 @@
 				this.updateUrlParam();
 			},
             updateUrlParam() {
-                this.setFilterValue("selected_items", this.checked.join('|'), true);
+                this.setFilterValue("selected_items", this.checked.join('|'));
             },
             onDropdownOpen() {
                 this.fetchLayerData();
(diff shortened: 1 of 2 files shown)


visualizations/frontend/filters/v2/TMultiSelector.vue
```diff
@@ -150,13 +150,16 @@
...
+        },
 		methods: {
             onVisualizationInit() {
 				const initial_value = this.getFilterValue("selected_items");
 
                 if (initial_value) {
                     this.checked = initial_value.split('|');
+                } else if (this.defaultValue) {
+                    this.checked = this.defaultValue.split('|');
+                    this.setFilterValue("selected_items", this.defaultValue);
```
sareg0, line 162: Would it make sense to `setFilterValue` and then read from it for `this.checked`? Rather than from the `defaultValue` prop?

Not sure it makes sense, but sort of like this. 

```
                if (this.defaultValue) {
                    this.setFilterValue("selected_items", this.defaultValue);
                }
                    
                this.checked =  this.getFilterValue("selected_items");
```

Please feel free to correct me if my understanding of the `checked`/`selected_items` nexus is incorrect.

sareg0 (commented):
Looks good. Only a minor question.

sareg0 (approved):

visualizations/frontend/filters/v2/TMultiSelector.vue
egcastro: We need to read first from `filters` to give priority to URL params before setting the default value.
This will be the order: `URL query param > default-value > empty`.
That's mainly the reason we can't call `setFilterValue` if `defaultValue` is set and then read only once because we'll be giving priority to default value even if there was already a value set by the URL.

sareg0: Got it! Thanks for the explanation!

PR #108, opened 2022-08-08
Added front-end modify columns code, including update to TTable and TSelectColumns

msorber:
[Pitch](https://www.notion.so/snyk/Table-support-for-modifying-columns-54daf4c649894ff8904e11a175e03312)

Demo: https://user-images.githubusercontent.com/78518576/183450871-294edc61-68ce-4c96-9ec3-525ccb393973.mov

### To test: 

1. in the topcoat IDE check out the modify_columns branch of snyk-insights
2. create a library component called “TestCsvExport.vue” and copy the contents of [TCsvExport](https://github.com/topcoat-data/topcoat-public/blob/851ed2a2f6c9cccd495599c248bd4cad2ec9e779/library/components/TCsvExport.vue) into it.
3. create a visualization called “TSelectColumns.vue” and copy the contents from the [TSelectColumns.vue](https://github.com/topcoat-data/topcoat-public/blob/851ed2a2f6c9cccd495599c248bd4cad2ec9e779/visualizations/frontend/other/TSelectColumns.vue) component in this branch.
4. create a visualization called “TestTable.vue”
- copy the contents of TTable
- update TCsvExport -> TestCsvExport
- update the references to “TTable” in issue_detail.html to “TestTable”
- update the references to “TTable” in issue_pdf_template to “TestTable”
5. Talk to me when it doesn’t work and I’ll figure out what step(s) I missed :joy:

### Usage
There is a new prop on the table. that accepts an array of column configurations in the following format:
:modifiableColumns="[
                {
                    displayColumn: 'SCORE',
                    layerColumns: ['PRIORITY_SCORE'],
                    displayByDefault: true,
                },
                {
                    displayColumn: 'CVE',
                    layerColumns: ['CVE'],
                    displayByDefault: true,
                },
            ]"
Where displayColumn matches the header of the column, layerColumns are the columns(s) in the layer file that the display column relies on, and an optional displayByDefault.

Additionally in the table's layer, each sql column that can be filtered needs to be wrapped in a check:
{% if filter('column_list') | selected_column('PROBLEM_TITLE') %}
	PROBLEM_TITLE,
{% endif %}

Optionally, if there is a join to retrieve columns that won't be used, that join can be averted:
{% if filter('column_list') %}
    {% if filter('column_list') | selected_column(['project_id', 'project_name']) %}
        left join "STITCH_INGESTION"."POSTGRESQL__PROD__SNOWFLAKE"."PROJECTS" p
        on i.project_id = p.id
    {% endif %}
{% else %}
    left join "STITCH_INGESTION"."POSTGRESQL__PROD__SNOWFLAKE"."PROJECTS" p
    on i.project_id = p.id
{% endif %}


sareg0 (commented):
Can I trouble you to give some more context on the PR description about the work. For example 
* Link to the pitch or the design
* Screen recording or screenshots of the end product.
* Instructions on how to test
Many thanks in advance 🙏

library/components/tableComponents/TSelectColumns.vue
```diff
@@ -0,0 +1,173 @@
...
+                :value="item"
+                v-model="checked"
+              />
+            </div>
+          </li>
+          <small v-if="!filterableColumns.length"> No items found </small>
+        </ul>
+      </div>
+    </div>
+  </t-dropdown>
```
sareg0, line 96: Was t-multi-selector not usable for this use case?

msorber:
> 

My apologies, updated the description with pitch, demo, and testing instructions. :D

library/components/tableComponents/TSelectColumns.vue
msorber: I could have used t-multi-selector, but only by adding a ton of special-case code to it that would have made it much harder to reason about and edit in the future. t-mulit-selector assumes the data to display is coming from a layer, but given that the data for this makes more sense to provide as a prop it was not a great fit under the hood. 😢

sareg0: ah, ok. Understood.

library/components/tableComponents/TSelectColumns.vue
```diff
@@ -0,0 +1,76 @@
...
+      <t-loading-spinner v-if="is_loading" position="relative" />
+      <download-outline-icon v-else :size="18" />
+      {{ label }}
+    </div>
+  </button>
+</template>
+
+<script>
+export default {
+  name: "TCsvExport",
```
sareg0, line 26: Does the file name need to be updated to match the component name?

library/components/tableComponents/TSelectColumns.vue
```diff
@@ -0,0 +1,76 @@
...
+      border-1 border-[#C3C2CB]
+      text-[#555463]
+      bg-white
+      w-max
+      p-1
+    "
+    style="height: fit-content"
+    :disabled="is_loading"
+    @click="download"
+  >
```
sareg0, line 15: Is it possible to use `BaseButton`?

sareg0 (commented):
Thank you for the extra description. Unfortunately the demo video does not seem to work for me. 



> In a table, add prop "filterableColumns" that is an array of arrays:
> :filterableColumns="[['SCORE','PRIORITY_SCORE', true], ['ISSUE', ['ISSUE_TYPE', 'PROBLEM_TITLE']]]"
> 
> Each sub-array is a column that the user can show or hide with 2 or 3 parameters: the front-end column name in the TTable, a string with the sql columns that front-end column depends on or an array of sql columns if the front end column combines several, and an optional boolean that if it is true the column will default to being displayed.

I wonder if we could use a different the data structure for `filterableColumns`. An array of arrays, that depend on position seems like a hard data structure to document, and make sure people get right. Perhaps we could get Ezra's input on this syntax for `filterableColumns`.

Questions:
* Do we really need to specify the sql column and the display column? Is it possible to simplify to only needing the display column?  
* Do we need to give the user the ability to specify the filterable columns? Could we  make all columns modifiable and save on a big list. As the columns on a table grow maintaining that list could be quite a task: 
  * e.g. `canColumnsBeModified`, all columns can be modified
* Re. "optional boolean that if it is true the column will default to being displayed". Does this mean that even if a column is 'filtered out', if this optional boolean is present and true, it will still show the column?

msorber:
>

- Yes both display column and sql column are necessary right now. It would be possible to get the sql column for many of the columns from the display column in the future but that is more work than I have time for at the moment and won't cover the display columns that depend on multiple back end columns. The sql columns are necessary so layers can avoid joins and fetching unnecessary data which will improve performance. 
- Also yep, chatted with Ezra about that a while ago and not all columns should be filterable so we unfortunately can't just list them all and call it a day. 😢 
- Correct, it lets the user hide the column if they want to but assumes that they want to see it by default.

library/components/tableComponents/TSelectColumns.vue
msorber: Absolutely, will update :D

library/components/tableComponents/TSelectColumns.vue
```diff
@@ -0,0 +1,76 @@
...
+  data: () => ({
+    is_loading: false,
+  }),
+  methods: {
+    download() {
+      this.is_loading = true;
+      payload = {
+        layer: this.tLayer,
+      };
+      filters = this.getFiltersState ? this.getFiltersState : {};
```
egcastro, line 50: Is there any case were `getFiltersState` has returned something different than an object? It should not, but I'm interested into know if you have seen that happened

library/components/TCsvExport.vue
```diff
@@ -33,25 +33,25 @@ export default {
...
-      let filenameFilters = "";
-      for (let filterKey of urlParams.keys()) {
-        filters[filterKey] = urlParams.get(filterKey);
-        filenameFilters += "-" + filterKey + "_" + urlParams.get(filterKey);
-      }
       payload = {
-        filters,
         layer: this.tLayer,
       };
+      filters = this.getFiltersState ? this.getFiltersState : {}
```
egcastro, line 50: We might change the default/initial value for `filters` in the page store to be `{}` instead of `null`. Here: https://github.com/snyk/topcoat-core/blob/main/frontend/src/store/modules/page.js#L40

visualizations/frontend/other/TTable.vue
```diff
@@ -19,8 +18,25 @@
...
+
+        <div class="tableControls">
+          <!-- <div>Future Group By</div> -->
+          <t-select-columns
+            v-if="filterableColumns"
+            :filterableColumns="filterableColumns"
+            @updateFilteredColumns="updateFilteredColumns"
+            :t-layer="layer"
+          />
+          <TestCsvExport1
```
sareg0, line 34:
```suggestion
          <TCsvExport
```

library/components/tableComponents/TSelectColumns.vue
sareg0: oh, interesting. I see that this button has the csv download information in it, and I don't see any checkbox elements anymore? Was there something that went wrong during the renaming?

library/components/TCsvExport.vue
```diff
@@ -64,7 +64,7 @@ export default {
               encodeURIComponent(response.data.data)
           );
         }
-        let filename = this.tLayer + filenameFilters + ".csv";
+        let filename = this.tLayer + ".csv";
```
egcastro, line 67:
```suggestion
        const filename = `${this.tLayer}.csv`;
```

library/components/tableComponents/TSelectColumns.vue
msorber: I copied and pasted the csv update into the wrong file, fixed..... 😖

library/components/TCsvExport.vue
msorber: Is that something that I can just update now as part of this PR?

visualizations/frontend/other/TTable.vue
msorber: Already fixed :)

egcastro:
I think that the proposed API is quite complex to understand from a user perspective. Positional/order dependant parameters are not that user friendly most of the time.
This is the proposed API for the filterableColumns prop:

```js
filterableColumns = [
    ['SCORE','PRIORITY_SCORE', true],
    ['ISSUE', ['ISSUE_TYPE', 'PROBLEM_TITLE']]
]
```

The issues I see with this proposal are:
- positional parameters are hard to understand/grasp.
- lack of type consistency: mixed types in the middle parameter, string or array of strings

I like to treat component's props as our public API. I like to define props definition using Typescript types for better understanding of its shape, even if we are not using TS at the moment, I'd leave the type definition on the PR description.

And a easier to understand prop API proposal: 

```typescript
interface FilterableColumn {
    columnName: string,
    layerColumnNames: Array<string>,
    isDisplayed?: boolean,
}

type FilterableColumns = Array<FilterableColumn>;

filterableColumns = [
    {
        columnName: 'SCORE',
        layerColumnNames: ['PRIORITY_SCORE'],
        isDisplayed: true,
    },
    {
        columnName: 'ISSUE',
        layerColumnNames: ['ISSUE_TYPE', 'PROBLEM_TITLE'],
    }
]
```

_Why I think this flavour is easier to understand?_
Each filterable column has a consistent and named object shape, you will always find `columnName` and `layerColumnNames` on it and they both will be the same type, string and array of strings, no matter the amount of columns (1 to N).
This makes API easy to grasp, no order to be remember and mixed types that are error prone.
On the other hand it's easier to parse those. Instead of having a code that will have some magic indexes like:

```js
const COLUMN_NAME = 0;
const LAYER_COLUMN_NAMES = 1;
const IS_DISPLAYED = 2;

for (filter of filterableColumns) {
    const columnName = filter[COLUMN_NAME];
    const layerColumnNames = filter[LAYER_COLUMN_NAMES]
    const isDisplayed = filter[IS_DISPLAYED] || false;
}
```

We can have something like:

```js
for (filter of filterableColumns) {
    const {columnName, layerColumnNames, isDisplayed = filter;
}
```

What do you think about this proposal?

library/components/TCsvExport.vue
egcastro: Oh no, I did not mean to change it right now, it was a thought while watching you have to do this validation. We can change that on a separated task.

visualizations/frontend/other/TTable.vue
```diff
@@ -1067,11 +1108,35 @@ export default {
...
+      let allValidColumns = [];
+      filterableColumnsToShow.forEach((fcts) => {
+        allValidColumns = allValidColumns.concat(fcts.sqlColumns);
+      });
+      // remove duplicate entries
+      allValidColumns = [...new Set(allValidColumns)];
+      if(allValidColumns.includes(undefined)){
+        console.error('Modify Column Configuration includes "undefined" value, please check that all sql column names are quoted!')
+      }
+      thisTable.filters.column_list = JSON.stringify(allValidColumns);
```
egcastro, line 1123: This is breaking store immutability. We should create a layer mutation to add `column_list` to filters or you should set the whole `filters` object extending it with what you need to add (column_list) using the `storeAttribute` mutation from the layer store.

egcastro:
I did not manage to test the PR properly. I can't make it run locally. I followed the steps in the description but I'm not able to see `Modify columns` dropdown. I see two `Download CSV` instead.

image

Any clue what can I be doing wrong? 

This is the table:
image

here is the layer:
image

library/components/TCsvExport.vue
msorber: Awesome

msorber:
> 

Nothing, I goofed and copied a csv update into the TSelectColumns component

msorber:
> 

Agreed entirely that an array of params can get confusing, however it does have the non-trivial advantage of brevity which is why I went with it over an array of objects with consistent fields. Let me get Seth/Ezra/Kennan's input on their preference.

sareg0:
>     * Correct, it lets the user hide the column if they want to but assumes that they want to see it by default.

Isn't the default of a column to display anyway, if this property was not provided? Don't we use 'excludeColumns' to exclude ones we don't want to see?

msorber:
> > ```
> > * Correct, it lets the user hide the column if they want to but assumes that they want to see it by default.
> > ```
> 
> Isn't the default of a column to display anyway, if this property was not provided? Don't we use 'excludeColumns' to exclude ones we don't want to see?

excludeColumns is for columns that should never be shown on their own, for example a url that is combined with a description in the description's display column to make a clickable link with user-friendly text. The URL needs to be in the layer but since it is already being incorporated into another column it should never end up as a display column by itself.

The displayByDefault here is for columns that the user has the option to show or hide- and either case is valid but the developer wants to default to showing it.

visualizations/frontend/other/TTable.vue
```diff
@@ -1067,11 +1102,34 @@ export default {
...
+      filterableColumnsToShow.forEach((fcts) => {
+        allValidColumns = allValidColumns.concat(fcts.sqlColumns);
+      });
+      // remove duplicate entries
+      allValidColumns = [...new Set(allValidColumns)];
+      if(allValidColumns.includes(undefined)){
+        console.error('Modify Column Configuration includes "undefined" value, please check that all layer column names are quoted!')
+      }
+      const columnList = JSON.stringify(allValidColumns)
+      thisTable.filters.column_list = columnList;
```
egcastro, line 1118: I left a comment previously that might have been deleted  by GitHub so I'm posting it back:
This is breaking store immutability. We should create a layer mutation to add column_list to filters or you should set the whole filters object extending it with what you need to add (column_list) using the storeAttribute mutation from the layer store.

visualizations/frontend/other/TTable.vue
```diff
@@ -307,10 +330,14 @@ export default {
...
       pagerResetFunction: null,
-      internalColumns: [],
+      filterableColumnsToShow: [],
+      additionalFilters: {},
     };
   },
   computed: {
+      isPdf(){
+          return window.location.href.includes('/pdf?')
+      },
```
sareg0, line 340: How do you feel about making this a prop instead of a computed property?

Relying a path on the window url makes the implementation tied to the way the PDF is currently implemented. Which makes it harder to update the implementation of pdf, without affecting this component. I also wonder how this might play with the iframe? Maybe Emi could shed some light here

If the CSV export and the table are very coupled by design (and we can't separate them for now), surfacing a prop (`canDownloadCSV`, for example) makes what the component will display more explicit and independently controlled at the point of use. We can default it to 'false', for the PDF implementation not to break.

egcastro (approved):
LGTM! 🚀

PR #109, opened 2022-08-09
Tags filter and fixes

gau94rav:
What's new
- Tags filter from Product CL.

Fixes:
- Some product CL filters do not work well with `v-show` or `v-if` when used inside a dropdown, used `height: 0px` logic for them instead, which is working great.
- Removed un-used code.

Tags Syntax:
```
<t-tags-input
    t-layer="filter_cve"
    item-label="CVE"
    t-filter:selected_items="cve"
    :is-label-capitalized="true"
    :is-single-tag="true" // If only one tag, use this.
/> 
```

### DIFF
--- a/library/components/TDropdown.vue
+++ b/library/components/TDropdown.vue
@@ -1,8 +1,6 @@
 <template>
 	<div
-		class="relative font-sans cursor-pointer w-max dropdown-filter"
-		@mouseover="setActiveState"
-		@mouseleave="setActiveState($event, false)"
+		class="relative w-auto font-sans cursor-pointer dropdown-filter"
 		ref="dropdownFilter"
 	>
         <div
@@ -13,12 +11,11 @@
         </div>
 		<slot name="outside"></slot>
 		<div
-			v-show="isPopupOpen"
 			ref="popup"
-			class="absolute mt-1 bg-white rounded-lg shadow-md base-dropdown-menu w-max min-w-"
-			:class="alignClass"
+			class="absolute mt-1 bg-white rounded-lg shadow-md base-dropdown-menu w-max"
+			:class="[alignClass, isPopupOpen ? 'h-max': 'h-[0px] opacity-0 overflow-y-hidden']"
 			:style="{ zIndex }"
-            :key="componentKey"
+			@click.stop
 		>
             <slot></slot>
 		</div>
@@ -29,10 +26,6 @@
 	export default {
         props: {
             isActive: {
-                type: Boolean,
-                default: false,
-            },
-			isPersisted: {
                 type: Boolean,
                 default: false,
             },
@@ -48,7 +41,6 @@
 		data: () => ({
 			alignClass: '',
 			isPopupOpen: false,
-            componentKey: '',
 		}),
         computed: {
             activeClass() {
@@ -93,11 +85,6 @@
 				this.$nextTick(() => {
 					this.alignPopup();
 				})
-			},
-			setActiveState(e, state = true) {
-				if (!state && this.isPopupOpen) {
-					return;
-				}
 			},
 			openPopup() {
 				if (this.isPopupOpen) {
--- a/visualizations/frontend/filters/v2/TTagsInput.vue
+++ b/visualizations/frontend/filters/v2/TTagsInput.vue
@@ -1,0 +1,145 @@
+<template>
+	<div class="t-tags-input max-w-max min-w-full min-h-[43px] relative">
+		<div class="pb-2 font-bold tracking-widest" v-if="label">
+			{{ label.toUpperCase() }}
+		</div>
+		<base-multiple-value-input
+			v-if="isSingleTag"
+			v-model="selectedTags"
+			selection-only
+			:data="tags.length ? tags : tempTags"
+			:loading="loading"
+			:placeholder="placeholder"
+			@update="handleUpdate"
+			display-key="value"
+			value-key="value"
+			@opened="handleOpen"
+		/>
+		<base-tags-input
+			v-else
+			v-model="selectedTags"
+			selection-only
+			:data="tags"
+			:loading="loading"
+			:placeholder="placeholder"
+			@update="handleUpdate"
+			@opened="handleOpen"
+		/>
+	</div>
+</template>
+
+<script>
+	export default {
+		props: {
+			label: {
+				type: String,
+				default: '',
+			},
+			tKeyColumn: {
+				type: String,
+				default: 'KEY',
+			},
+			tValueColumn: {
+				type: String,
+				default: 'VALUE',
+			},
+			isSingleTag: {
+				type: Boolean,
+				default: false,
+			},
+			placeholder: {
+				type: String,
+				default: '',
+			}
+		},
+		data: () => ({
+			is_filter: true,
+			selectedTags: [],
+			tempTags: [], // base-multiple-value-input does not show selected items if `tags` is empty because data is fetched on open.
+		}),
+		computed:{ 
+			tags() {
+				const items = [];
+
+				if (this.rows.length) {
+					for (let row of this.rows) {
+						const key = row[this.tKeyColumn];
+						const value = row[this.tValueColumn];
+						if (key && value) {
+							items.push({
+								key: key.value,
+								value: value.value,
+							})
+						}
+					}
+				}
+				return items;
+			},
+			firstKey() {
+				// Used for single tag
+				if (this.tags.length) {
+					return this.tags[0].key;
+				}
+				
+				if (this.tempTags) {
+					return this.tempTags[0].key;
+				}
+				
+				return null;
+			}
+		},
+		methods: {
+			onVisualizationInit() {
+				const initial_value = this.getFilterValue("selected_items");
+
+				if (initial_value) {
+					const selectedTags = [];
+                    const tempTags = [];
+					const urlTags = initial_value.split('|');
+					
+					for (let item of urlTags) {
+						const splitPair = item.split('_');
+						const key = splitPair.length > 1 ? splitPair[1] : splitPair[0];
+						const value = splitPair[0];
+
+						if (key && value) {
+							if (this.isSingleTag) {
+								selectedTags.push(value);
+								tempTags.push({ key, value })
+							} else {
+								selectedTags.push({ key, value });
+							}
+						}
+					}
+					
+				    this.tempTags = tempTags;
+				    this.selectedTags = selectedTags;
+				}
+			},
+			handleUpdate(items) {
+				const values = [];
+				for (let item of items) {
+					if (this.isSingleTag) {
+						if (!this.firstKey) {
+							break;
+						}
+						values.push(`${item}_${this.firstKey}`);
+					} else {
+						values.push(`${item.value}_${item.key}`);
+					}
+				}
+				this.setFilterValue("selected_items", values.join('|'));
+			},
+			handleOpen() {
+				this.fetchLayerData();
+			},
+		},
+	}
+</script>
+
+<style>
+	.t-tags-input .vue--multiple-value-input__body {
+		position: relative !important;
+		min-width: 100% !important;
+	}
+</style>
--- a/library/components/TExpandable.vue
+++ b/library/components/TExpandable.vue
@@ -14,7 +14,7 @@
                 <slot name="icon"></slot>
             </div>
         </div>
-        <div class="px-2 pt-1" v-show="expanded">
+        <div class="px-2 pt-1" :class="expanded ? 'h-max' : 'h-[0px] opacity-0 overflow-y-hidden'">
             <slot :onExpandable="expanded"></slot>
         </div>
     </div>
--- a/library/templates/Report/ReportHeader.vue
+++ b/library/templates/Report/ReportHeader.vue
@@ -15,7 +15,7 @@
                     </span>
                 </div>
                 <t-dropdown disable-handle-class>
-                    <menu-down-icon slot="handle" class="relative right-1" />
+                    <menu-down-icon slot="handle" />
                     <div class="flex flex-col gap-2 px-2 py-2 md:px-5">
                         <a href="/logout">Logout</a>
                         <a href="https://snyk.io/" class="md:hidden">Go to snyk.io</a>
--- a/visualizations/frontend/filters/v2/TMultiSelector.vue
+++ b/visualizations/frontend/filters/v2/TMultiSelector.vue
@@ -166,7 +166,8 @@
                 if (this.checked.length === this.ids.length) {
                     this.checked = [];
                 } else {
-                    this.checked = this.ids;
+					// Only select what is visible, search can affect this.
+                    this.checked = this.menu.map(i => i.value);
                 }
 				this.updateUrlParam();
 			},


library/components/TDropdown.vue
```diff
@@ -1,8 +1,8 @@
 <template>
 	<div
 		class="relative font-sans cursor-pointer w-max dropdown-filter"
-		@mouseover="setActiveState"
-		@mouseleave="setActiveState($event, false)"
+		@mouseover="hasMouseOver = true"
```
gau94rav, line 4: When dropdown has child dropdown under, clicking on it is considered clicking outside which closes parent dropdown, this fix makes sure that does not happen.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,153 @@
...
+				}
+				this.setFilterValue("selected_items", urlValue, true);
+			}
+		},
+	}
+</script>
+
+<style>
+	.t-tags-input .vue--multiple-value-input__body {
+		@apply !relative !min-w-full;
```
sareg0, line 151: I looked this up, and Caniuse suggests it's not supported very well? 
https://caniuse.com/?search=%40apply

gau94rav: Oh I see, sure I can switch to regular css in that case, thanks @sareg0 .

gau94rav: Ahh I thought even tailwind dropped it's support 😛

sareg0: no no, it's ok. I am just new to tailwind, so it's a mistake on my part.

sareg0 (commented):
Is it possible to have separate PRs for the fixes for the child dropdowns and the addition of the tag filter.

gau94rav:
> Is it possible to separate the fixes for the child dropdowns and the addition of the tag filter?



> Is it possible to have separate PRs for the fixes for the child dropdowns and the addition of the tag filter.
> 
> In addition, is there a way we can test this out? Could you leave some guidance no testing in the PR?

Sure, though tags filter kind of depends on that dropdown fix to function properly inside `t-filter-group`

About testing, yes sure the way I tested this was by changing `revision` inside `config.yml` to this branch and then re-building modules by clicking on cloud icon inside IDE.

To test new dropdown, you can go inside `issues_detail.htmt` and change any component that uses `t-tags` to use `t-tags-input`.

library/components/TDropdown.vue
sareg0: Have you considered using event modifiers to overcome this bug? 

```
<div @click.self="doThat">...</div>
```
https://vuejs.org/guide/essentials/event-handling.html#event-modifiers


I only mention it because ProductCL also has to avoid closing the main dropdown when there are filters inside, and they have a recipe here that makes use of event modifiers to stop the event when necessary: https://productcl.netlify.app/?path=/docs/recipes-dropdownfilters--overview

I'm not sure if it solves your problem though.

I am

gau94rav: Woow, yup that seems to be working, let me test it a bit then I'll push it, thanks @sareg0 .

sareg0: NICE!

gau94rav: Alright seems to be working perfectly, thanks again, not sure why I never knew about it lol

gau94rav: Sadly we still need the height fix for a different issue by product cl, what happens is that if dropdown is invisible the t-tags filter fails to assign the exact selected items and shows +4 for all selected items instead.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+				default: '',
+			},
+			tKeyColumn: {
+				type: String,
+				default: 'KEY',
+			},
+			tValueColumn: {
+				type: String,
+				default: 'VALUE',
+			},
```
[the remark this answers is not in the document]
gau94rav: Most of tags layers have these names, I thought it's best to have these as defaults.

sareg0 (approved):
It looks great!

library/components/TDropdown.vue
```diff
@@ -13,12 +11,11 @@
         </div>
 		<slot name="outside"></slot>
 		<div
-			v-show="isPopupOpen"
 			ref="popup"
-			class="absolute mt-1 bg-white rounded-lg shadow-md base-dropdown-menu w-max min-w-"
-			:class="alignClass"
+			class="absolute mt-1 bg-white rounded-lg shadow-md base-dropdown-menu w-max"
+			:class="[alignClass, isPopupOpen ? 'h-max': 'h-[0px] opacity-0 overflow-y-hidden']"
```
egcastro, line 16: I think that this dynamic part is not working as expected, it's not producing any issues but it's not changing neither:
`isPopupOpen ? 'h-max': 'h-[0px] opacity-0 overflow-y-hidden']`




The one that is actually changing is the TExpandable.vue one

gau94rav: Yes this is just setting height of expandable section, elements still stay in DOM just height gets reduced and max on click.

`height: 0px` gets changed to `height: max-content`

Am I misunderstanding you question?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+					}
+				}
+				this.setFilterValue("selected_items", urlValue, true);
+			}
+		},
+	}
+</script>
+
+<style>
+	.t-tags-input .vue--multiple-value-input__body {
```
egcastro, line 150: This is more a question than a comment: where are this type of classes come from `.vue--multiple-value-input__body` ?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+					}
+				}
+				this.tempTags = tempTags;
+				this.selectedTags = selectedTags;
+			},
+			handleOpen() {
+				if (!this.initialized || this.canReloadOnOpen) {
+					this.fetchLayerData();
+					this.initialized = true;
+				}
```
egcastro, line 129: Why do we want to prevent the `fetching` to happen on every `open`?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+				return items;
+			},
+			firstKey() {
+				// Used for single tag
+				if (this.tags.length) {
+					return this.tags[0].key;
+				} else if (this.tempTags) {
+					return this.tempTags[0].key;
+				}
+				return null;
```
egcastro, line 90: I try to avoid this kind of `if else if` code structures, what do you think about this option:
```suggestion
				if (this.tags.length) {
					return this.tags[0].key;
				}
				
				if (this.tempTags) {
					return this.tempTags[0].key;
				}
				
				return null;
```

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+			}
+		},
+		keyColumn() {
```
egcastro, line 93: This method is not used anywhere, should we remove it?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+			return this.tKeyColumn ? this.tKeyColumn : this.findColumnByTag('keys');
+		},
+		valueColumn() {
```
egcastro, line 96: This method is not used anywhere, should we remove it?

library/components/TDropdown.vue
egcastro: yes, you have modified and add this behaviour to two components TExpandable and TDropdown. The one changing in the DOM is the TExpandable while the TDropdown is always hardcoded as `h-max` and it's not reacting to the `isPopupOpen` data change.
On the GIF I shared, the DOM node that is highlighted, correspond to the TDropdown, is not changing. The ones changing are the one from the TExapandable.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+				type: Boolean,
+				default: false,
+			},
+			placeholder: {
+				type: String,
+				default: '',
+			}
+		},
+		data: () => ({
+			is_filter: true,
```
sareg0, line 60:
```suggestion
			isFilter: true,
```

I don't see this being used anywhere though. Is it still needed?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+						const key = splitPair.length > 1 ? splitPair[1] : splitPair[0];
+						const value = splitPair[0];
+
+						if (key && value) {
+							if (this.isSingleTag) {
+								selectedTags.push(value);
+								tempTags.push({ key, value })
+							} else {
+								selectedTags.push({ key, value });
+							}
```
egcastro, line 118: Shouldn't we be assigning `tempTags` in both cases? In both cases we are binding data like this:
```js
    :data="tags.length ? tags : tempTags"
```
I guess, but I may be missing something, that `tempTags` is used when there is not `tags`, and `tags` are missing when there is no `rows` (fetched data). So I guess that `tempTags` should be assigned in both cases.

library/components/TDropdown.vue
gau94rav: Ahh I see what you mean, yes the t-dropdown only changes when you first time open the filters dropdown, that is a wrapper container.

While t-expandable are child component inside the t-dropdown.

https://user-images.githubusercontent.com/43262405/183873267-6fa1d440-c504-4871-92ea-bb3f1b7cc25e.mp4

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: This is coming from `BaseRangeSlider` component used inside `TTagsInput` which is part of Product CL.

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: Usually cve lists are used for tags filter and they are huge, might not be needed to fetch them again, but if that does not make sense I can remove this.

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: Got it, thanks.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,154 @@
...
+					if (this.isSingleTag) {
+						if (!this.firstKey) break;
+						let urlString = `${item}_${this.firstKey}|`
+						urlValue += urlString;
+					} else {
+						let urlString = `${item.value}_${item.key}|`
+						urlValue += urlString;
+					}
+				}
+				this.setFilterValue("selected_items", urlValue, true);
```
egcastro, line 143: The third parameter has been removed on the async filters PR. It was used to `notify` other components about the filter set but we do not need that flag anymore since we are relying on Vuex.
```suggestion
				this.setFilterValue("selected_items", urlValue);
```

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: This is part of topcoat-core actually, it's used in visualization_mixin, this determines whether a vis needs the height wrapper logic (required by plotly).

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: Temp tags takes values from url as `tags` on initialization, this hack was only required because the tags component by product cl does not show selected items until it has a array of objects available as `data`, so until we actually click on filter and load, nothing appears as selected.

egcastro:
Do you know if we can set a key by prop and avoid this selection:
image
I think that we always have only one `key` that is the same as the parent opened dropdown label (CVE) in this case

gau94rav:
> Do you

Yes if you use a prop called `is-single-tag` to switch it into only using one key. So if your tag layer only has one tag you can use this prop.

visualizations/frontend/filters/v2/TTagsInput.vue
egcastro: Yes but we are only applying this hack to `singleTag` components, is that right?
I'm asking because we are doing the same for `data` in both components:
```vue
		<base-multiple-value-input
			v-if="isSingleTag"
			v-model="selectedTags"
			selection-only
			:data="tags.length ? tags : tempTags"
			:loading="loading"
			:placeholder="placeholder"
			@opened="handleOpen"
			@update="handleUpdate"
			display-key="value"
			value-key="value"
		/>
		<base-tags-input
			v-else
			v-model="selectedTags"
			selection-only
			:data="tags.length ? tags : tempTags"
			:loading="loading"
			:placeholder="placeholder"
			@opened="handleOpen"
			@update="handleUpdate"
		/>
```

gau94rav: Ahh I see, my bad, checking again.

visualizations/frontend/filters/v2/TTagsInput.vue
sareg0: I'd say we could remove this and look for optimization opportunities once we see how it behaves in production.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -0,0 +1,153 @@
...
+					if (this.isSingleTag) {
+						if (!this.firstKey) break;
+						let urlString = `${item}_${this.firstKey}|`
+						urlValue += urlString;
+					} else {
+						let urlString = `${item.value}_${item.key}|`
+						urlValue += urlString;
+					}
+				}
+				this.setFilterValue("selected_items", urlValue);
```
egcastro, line 129: I'm not quite sure of the format we chose to store tags.
`cve=<value>_<key>|<value>_<key>|<value>_<key>|`
Example:
`cve=ALAS-2018-1010_CVE%257CALAS-2018-1026_CVE%257C`
decoded:
`cve=ALAS-2018-1010_CVE|ALAS-2018-1026_CVE|`

gau94rav: Yes this was something that was hard to understand back when we used it for the first time, the reason for this is because of how it is used inside layers.

Layers use filters to compare it like this.

WHERE key = {{ url_filter.key | to_sql_tags  }}
select value

basically `to_sql_tags` expects data in this format inside backend `value_key`, I am sure that backend method can be made better.

sareg0:
> Yes if you use a prop called `is-single-tag` to switch it into only using one key. So if your tag layer only has one tag you can use this prop.

Does it make sense to split single tag and multi tag inputs? 

I feel like that would keep the props API clean. The draw back might be repeated logic across the two.

gau94rav:
> > Yes if you use a prop called `is-single-tag` to switch it into only using one key. So if your tag layer only has one tag you can use this prop.
> 
> How do do you splitting the single tag and multi tag inputs into separate components?
> 
> I feel like that would keep the props API leaner. The draw back might be repeated logic across the two.

Product CL has two components for tags
1. BaseTagsInput
2. BaseMultipleValueInput

If `isSingleTag` is true then #2 is used else #1, the reason is because with #1 you will always have to select the key first then it's values, even if only one key is available in data.

Thankfully they both accept the data in same form, array of objects with key: val pairs, otherwise two different types of init logics might have been needed.

visualizations/frontend/filters/v2/TTagsInput.vue
egcastro: ok, on that case, I would suggest a small change on the implementation to make this easier. Instead of do some string addition, we can push all the values to an array and then use Array.join method to avoid the trailing `|`.
Example:
```js
const values = [];
for (let item of items) {
	if (this.isSingleTag) {
		if (!this.firstKey) {
		    break;
		}
		values.push(`${item}_${this.firstKey}`);
	} else {
		values.push(`${item.value}_${item.key}`);
	}
}
this.setFilterValue("selected_items", values.join('|'));
```

egcastro:
So in our case, where CVE is only and always will be the only key, we should skip that two step UI right?

gau94rav:
> So in our case, where CVE is only and always will be the only key, we should skip that two step UI right? I am asking, is not CVE always be the `key` in the CVE filter dropdown?

Well a single key layer should not even be inside tag filter maybe and just be a multi selector but maybe @eftanzer can help with this a bit more, because currently cve layer and layers subscribed to it are expecting it in layer format.

egcastro:
I think that we are misunderstanding each other 😄 
I'm referring to the UI component only not the layer behind it and the `{value, key}` filter format.
What I'm trying to say is if there any way to skip the step of selecting the CVE key in the dropdown, because it will always be CVE and only CVE for that dropdown.
I saw that we are fetching the key from the backend, maybe this is the reason we can't skip this step 🤷‍♂️

gau94rav:
> I think that we are misunderstanding each other 😄 I'm referring to the UI component only not the layer behind it and the `{value, key}` filter format. What I'm trying to say is if there any way to skip the step of selecting the CVE key in the dropdown, because it will always be CVE and only CVE for that dropdown. I saw that we are fetching the key from the backend, maybe this is the reason we can't skip this step 🤷‍♂️

Oh my bad again haha, actually you can skip that using `is-single-tag` prop.

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: @egcastro only `base-multiple-value-input` needs `tempTags`, because it cannot show selected properties until it has array of objects as data but `base-tags-input` does not need it, this is a weird limitation in base component, also temp tags should not be assigned as main variable for either of these, it does not get updated after initial call, as soon as `tags` gets data from backend, that should be used instead.

egcastro (approved):
LGTM! 🚀

PR #110, opened 2022-08-09
fix: Use value for getColumn ids

egcastro:
Use `value` instead of `rendered` attribute for `ids`.
Results comes with two attributes: rendered and value.
`value` is the raw value
`rendered` is a formatted version of the `value`.

Examples:
image

Related PR: https://github.com/snyk/topcoat-core/pull/487

### DIFF
--- a/visualizations/frontend/filters/v2/TMultiSelector.vue
+++ b/visualizations/frontend/filters/v2/TMultiSelector.vue
@@ -133,12 +133,16 @@
 
                 if (values && titles) {
                     for (let index in values) {
-                        const value = values[index];
+                        const value = values[index] && values[index].toString();
                         const title = titles[index];
+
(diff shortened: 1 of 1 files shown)


visualizations/frontend/filters/v2/TMultiSelector.vue
```diff
@@ -124,7 +124,7 @@
             },
             ids() {
 				const column_name = this.tKeyColumn ? this.tKeyColumn : this.findColumnByTag('ids');
-				return this.getColumn(column_name);
+				return this.getColumn(column_name, 'value');
```
gau94rav, line 127: I see, rendered has special string formatting applied to it which results in breaking of url, thanks this makes sense.

gau94rav (approved):
LGTM

PR #111, opened 2022-08-09
Fixed the "no data" message not showing up when filters reduced the # of matching rows to 0

msorber:
Bug: https://snyksec.atlassian.net/browse/DP-241

Note: there were actually 2 distinct issues:
1.  When filters reduce the total # of rows to 0 the spinner kept spinning, the currently shown rows stayed visible, and the "no data" message did not appear
2. When the url filters resulted in no rows when the page is loaded the table didn't show up at all.

This PR addressed both.

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -385,12 +385,6 @@ export default {
   },
   watch: {
     loading() {
-      if (!this.loading) {
-        this.updateEndIndex(this.rowsPerPage);
-        this.updateStartIndex(0);
-        this.pagerResetFunction();
-        this.fetchTotalRows();
-      }
       this.showSpinner = this.loading;
     },
     internalSelectedItem() {
@@ -406,8 +400,13 @@ export default {
     this.fetchTotalRows();
     this.setInternalColumns();
   },
-
   methods: {
+    onVisualizationUpdated() {
+      this.updateEndIndex(this.rowsPerPage);
+      this.updateStartIndex(0);
+      this.pagerResetFunction();
+      this.fetchTotalRows();
+    },
     // internal columns takes any column configurations from the TColumnConfig
     // component and creates default column configurations based on the fields
     // of the first row of data. That way users don't have to manually specify
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -12,8 +12,8 @@
       @updateSearchTerm="updateSearchTerm"
     />
 
+
     <div
-      v-if="isDataAvailable"
       id="tableContainer"
       ref="tableContainer"
       :style="columnWidthsStyle"
@@ -22,6 +22,14 @@
       <div v-if="title" id="title" class="spanAllColumns">
         {{ title }}
       </div>
+
+<!-- Special case when there are no rows, so there are no column headers, 
+so the no data message below can't show up -->
+        <!-- <div v-if="!isDataAvailable">
+            <div><i class="i-fa-solid i-fa-inbox"></i></div>
+            Top: <div>{{ noDataMessage }}</div>
+        </div> -->
+
 
       <!-- Empty div to keep the headers lined up with their columns when there are exapnd/collapse buttons  -->
       <div v-if="canCollapseDetailRows"></div>
@@ -312,6 +320,7 @@ export default {
       showSpinner: true,
       displayRows: [],
       pagerResetFunction: null,
+      internalColumns: [],
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -380,6 +376,38 @@ export default {
...
     },
+    layerRows() {
+      if (typeof this.rows === "function")
+        return this.rows(this.$attrs?.["t-layer"]);
+      if (this.rows) return this.rows;
+      return [];
+    },
+  },
+  watch: {
+    loading() {
```
sareg0, line 387: This might be a super silly question. 

What 'loading' are we watching? I don't see a data property called 'loading'.

egcastro, line 387: There is already a callback for this case, you can see it here: https://github.com/snyk/topcoat-core/blob/main/frontend/src/mixins/visualization_mixin.js#L425-L432

Basically, we should be able to move all the code inside the `if (!this.loading)` into the `onVisualizationUpdated` method.

visualizations/frontend/other/TTable.vue
```diff
@@ -380,6 +376,38 @@ export default {
...
+      if (!this.loading) {
+        this.updateEndIndex(this.rowsPerPage);
+        this.updateStartIndex(0);
+        this.pagerResetFunction();
+        this.fetchTotalRows();
+      }
+      this.showSpinner = this.loading;
```
egcastro, line 394: and `showSpinner` could be converted into a `computed` value I guess, or you can rely on `this.loading` instead of creating a new state.

visualizations/frontend/other/TTable.vue
msorber: That is not a silly question at all, it is provided by the back end with a [mixin](https://v2.vuejs.org/v2/guide/mixins.html?redirect=true). Looks like it is a computed property in layer_mixin.js but I haven't tested that.

msorber: Sweet! That is way simpler!

visualizations/frontend/other/TTable.vue
msorber: I think at one point it didn't always match this.loading, let me check if that is still the case.

msorber: No go, 😢 when the TTable fetches a new page of data "loading" isn't set to true so the spinner wouldn't show without the showSpinner state. We should probably revisit how fetching new pages is handled at some point to take advantage of your updates, and eliminate the need for showSpinner, but I think that is out of scope.

visualizations/frontend/other/TTable.vue
```diff
@@ -380,6 +376,38 @@ export default {
...
+    },
+    internalSelectedItem() {
+      this.$emit("update:selectedItem", this.internalSelectedItem);
+      this.$emit("selectedItemChanged", this.internalSelectedItem);
```
egcastro, line 392: Why do we need to emit two events that looks the same?

visualizations/frontend/other/TTable.vue
```diff
@@ -380,6 +376,38 @@ export default {
     uniqueOriginalRows() {
       return _.uniq(this.internalRows.map((r) => r.originalRow));
     },
+    layerRows() {
+      if (typeof this.rows === "function")
+        return this.rows(this.$attrs?.["t-layer"]);
+      if (this.rows) return this.rows;
+      return [];
```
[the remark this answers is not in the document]
msorber: Consistent conventions are way more important than my slight preference for being so lazy as to avoid 2 keystrokes. :joy: Updated.

visualizations/frontend/other/TTable.vue
msorber: This is from pre-topcoat code. The original table is meant to be very flexible, so most users will want to use the .sync modifier that will work with the first line, but there are times when that won't work/be sufficient so users can also react to the event instead.

visualizations/frontend/other/TTable.vue
msorber: Yep, I removed the watcher

visualizations/frontend/other/TTable.vue
sareg0: oh mixins. The trick me every time.

msorber: same... I love the idea but magic functionality you can't see? Hard to reason about 🙄

egcastro:
PR is working properly, the only thing I'd like to improve (creating a ticket for later) is the design of the `no data` state:

Screenshot 2022-08-12 at 09 30 33

Something that I notice while testing the PR is that when you end in the `No data` state while clicking and adding filters, you get the image from above. Then if you refresh the page with those filters in the URL, you get this `No data` + `Spinner` state:

Screenshot 2022-08-12 at 09 31 00

And more important, I found an issue not related with this PR but with table data fetching overall that I will try to fix soon to improve Table data fetching performance!

egcastro (approved):
LGTM!

sareg0 (approved):

PR #112, opened 2022-08-12
Feat/improved filters

gau94rav:
This PR adds two new filters.
1. Range Input Slider
2. Date Picker

Using a new date picker will help us get rid of Ant Design from backend as well.

**Syntax**
Range Input Slider
```
        <t-range-input
            t-layer="slider_layer"
            t-filter:min="min_val"
            t-filter:max="max_val"
            t-min-column="MIN"   // From sql file
            t-max-column="MAX" // From sql file
        />
```

Date picker (single): Skip end_date
```
        <t-date-pick
            t-layer="single_picker_layer"
            t-filter:start_date="date"
            t-filter:date_preset="singe_preset"
        />
```

Date picker (range)
```
        <t-date-pick
            t-layer="range_picker_layer"
            t-filter:start_date="start_date"
            t-filter:end_date="end_date"
            t-filter:date_preset="range_preset"
        />

```

Depends on this backend [PR](https://github.com/snyk/topcoat-core/pull/489).

Sample for both date and range slider filters.
Screenshot 2022-08-12 at 6 19 36 PM


sareg0 (commented):
Could you leave us some notes on how to test the components? Do you have a branch you have been testing them out on?

Are there tasks this is linked to, or a pitch? Could you link those tickets. Mainly so I can understand why we are replacing the calendar components we already have.

gau94rav:
> Could you leave us some notes on how to test the components? Do you have a branch you have been testing them out on?

Hi @sareg0 , you can only test them locally for now, because both these filters depend on backend PR as well.

To test
1. Switch to `feat/improved-filters` branch in topcoat-core on local.
2. `cd frontend && npm install && npm run dev`
3. Open IDE, place my sample syntax, thankfully you can just leave your layer files empty, unless you want to specify persist filters.
4. Inside config.yml switch revision to `feat/improved-filters` and click build modules button in sidebar.

gau94rav:
> > Could you leave us some notes on how to test the components? Do you have a branch you have been testing them out on?
> 
> Hi @sareg0 , you can only test them locally for now, because both these filters depend on backend PR as well.
> 
> To test
> 
> 1. Switch to `feat/improved-filters` branch in topcoat-core on local.
> 2. `cd frontend && npm install && npm run dev`
> 3. Open IDE, place my sample syntax, thankfully you can just leave your layer files empty, unless you want to specify persist filters.
> 4. Inside config.yml switch revision to `feat/improved-filters` and click build modules button in sidebar.

@sareg0 One correction for range slider, it can be used with a layer file and only expects two values from it, a min and a max number.

sample

`SELECT 20 AS min, 60 as max;`

Now that I think about it, it should also be able to do that with just prop, will push that change next time.

visualizations/frontend/filters/v2/TRangeInput.vue
```diff
@@ -0,0 +1,108 @@
+<template>
```
sareg0, line 1: Maybe we should call it `TRangeSlider`? Just to match the naming inside the file. Also, does Ezra have a preference on what the component should be called?

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,282 @@
...
+            <div class="presets-dropown" slot="sidebar">
+                <div
+                    v-for="preset in presets[mode]"
+                    :key="preset.key"
+                    :style="{ color: preset.key === selectedPreset.key ? '#1284e7' : '#73879c' }"
+                    class="cursor-pointer"
+                    @click="handlePreset(preset)"
+                >
+                    {{ preset.label }}
+                </div>
```
sareg0, line 27: could we make these buttons or other focusable elements? I think it is better practice for clickable elements.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,282 @@
...
+        },
+        handleClear() {
+            this.date = [];
+            this.selectedPreset = {};
+            this.unsetFilterValue("start_date");
+            this.unsetFilterValue("end_date");
+            this.unsetFilterValue("date_preset");
+            this.unsetFilterValue("date");
+        },
+        formatDate(dateValue) {
```
sareg0, line 231: Do we have access to `moment` here? It might be helpful to use it for formatting dates.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,282 @@
...
+                } else if (item === "MM") {
+                    date.push(month);
+                } else if (item === "DD") {
+                    date.push(day);
+                } else {
+                    // It's a seperator
+                    date.push(item);
+                }
+            }
+            return date.join("");
```
sareg0, line 256: Do you have access to `moment` here? It might be helpful to use it for formatting dates. 

```suggestion
    return moment().format('YYYY-MM-DD')
```

visualizations/frontend/filters/v2/TRangeInput.vue
gau94rav: Yes, there is already a old and broken component called that, for now I chose a different name so that it will be easy to review, once this PR gets approved I'll rename it to `TRangeSlider`

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: Yes, I can access moment here but @egcastro had some opinion on moment I think, basically moment library is bloated and I think is no longer getting updates, Emi can confirm.

sareg0: I presume that Emi was going to suggest using `date-fns`. From what I have seen, `date-fns` has a very similiar API to moment. However It's not a problem to format the dates by hand if you want to avoid dependencies.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,282 @@
...
+                >
+                    {{ preset.label }}
+                </div>
+            </div>
+
+            <div slot="icon-clear"></div>
+            <div slot="icon-calendar">
+                <div class="flex items-center gap-1">
+                    <reload-icon @click="handleReload" class="cursor-pointer" />
+                    <close-icon @click="handleClear" class="cursor-pointer" />
```
sareg0, line 34: Could we use a button here? This would make the element focusable and would remove the need for changing the cursor.

visualizations/frontend/filters/v2/TDatePicker.vue
sareg0, line 27: Could we use a button here? Same remark as below

visualizations/frontend/filters/v2/TRangeInput.vue
sareg0: oh I see. Is the rollout plan to Implement these components, test them and then remove the old components?

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,282 @@
...
+            }
+            this.date = date;
+        },
+        handleClear() {
+            this.date = [];
+            this.selectedPreset = {};
+            this.unsetFilterValue("start_date");
+            this.unsetFilterValue("end_date");
+            this.unsetFilterValue("date_preset");
+            this.unsetFilterValue("date");
```
sareg0, line 229: Do you think there is the possibility to move to 'start_date' and 'end_date' for the single date selection as well? It might simplify the component a bit and also remove the possibility of having a url with `date`, `start_date`, and `end_date` where it's hard to tell what's taking precedence.

example:
`start_date=2022-08-01&end_date=2022-08-01`

visualizations/frontend/filters/v2/TDatePicker.vue
egcastro: Exactly, we can use moment.js on this PR but track it so we can get rid of it when we switch over to `date-fns`

visualizations/frontend/filters/v2/TRangeInput.vue
gau94rav: Absolutely

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: Yes, this is what I thought first as well, it can be as simple as, if there is end_date then it's range otherwise it's a single calendar.
I say we go for it, we can still discuss more about presets for future changes but this is trivial.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,263 @@
...
+
+            <div slot="icon-clear"></div>
+            <div slot="icon-calendar">
+                <div class="flex items-center gap-1">
+                    <button>
+                        <reload-icon @click="handleReload" />
+                    </button>
+                    <button>
+                        <close-icon @click="handleClear" />
+                    </button>
```
sareg0, line 38: let's move the 'click' event to the buttons. That way the whole element is clickable, not just to the icon.

sareg0 (approved):
I didn't get the chance to test them out locally, however I'd be happy to give them a run through in snyk insights once they're merged.

library/components/TDropdown.vue
```diff
@@ -52,6 +56,12 @@
         },
 		mounted() {
 			onClickOutside(this.$refs.dropdownFilter, (event) => {
+				// Temporary solution
+                // If a component is appending it's inner dropdown directly to body.
+                for (let cl of this.outsideElementClasses) {
+                    // If an outside element is still open, keep the dropdown from closing.
+                    if (document.querySelector(`.${cl}`)) return;
+                }
```
sareg0, line 64: Did the calendar stay open, when the dropdown closed?

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,275 @@
...
+                >
+                    {{ preset.label }}
+                </button>
+            </div>
+
+            <div slot="icon-clear"></div>
+            <div slot="icon-calendar">
+                <div class="flex items-center gap-1">
+                    <button @click="handleClear">
+                        <close-icon />
```
sareg0, line 34: I'm not sure if anyone has already picked this up, but clicking the 'x' icon opens the calendar, and clicking the calendar icon does nothing. 


https://user-images.githubusercontent.com/1307818/186391489-e5211533-8773-4d7d-8450-84c195b7cf5e.mov

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -0,0 +1,275 @@
+<template>
+    <div class="relative flex t-date-picker">
+        <!-- Date Picker -->
+        <base-date-picker
+            read-only
+            show-clear-button
```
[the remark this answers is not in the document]
gau94rav: Missed it, thanks for finding, this was from a date picker I used before this one.

library/components/TDropdown.vue
gau94rav: When you click outside, it will first close calendar, then another click will close dropdown.

sareg0: That's with this change, but what about earlier? Why did you need this? Was it because the dropdown closed and the calendar stayed open?

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: > I'm not sure if anyone has already picked this up, but clicking the 'x' icon opens the calendar, and clicking the calendar icon does nothing. The pointer is a bit off in the recording, but I'm clicking the 'x' at first, and clicking the 'calendar' icon after.
> 
>  calendar-close-reload.mov

Strange but had to separate the loading icon for this to work, weird but works.

library/components/TDropdown.vue
gau94rav: Yes, the reason behind this is that (as I understand), this date picker is appended directly to body when it is opened, instead of being a child of dropdown, it is a direct second child of `<body />`, so with this fix we make sure that whatever classes we specify dropdown to watch before closing are checked before it is collapsed.

gau94rav: Because of this, @click.stop also does not work here.

sareg0: 😬 ah, ok. I was just curious. I can't think of a better solution right now, but I want to understand fully.

PR #113, opened 2022-08-12
un-removed required loading watcher

msorber:
I goofed when testing some cleanup for the fix for the “no data” message bug and removed a watcher that I thought was unnecessary but was actually needed. This re-adds the watcher.


sareg0:
What is going wrong? Why is it necessary?

PR #115, opened 2022-08-16
Add the "rendered" slot prop and updated the "value" slot prop to return the "value" property from the cell object

msorber:
Firefox does not support dates in the format month-day-year, fortunately the back end properly converts that to an iso compatible date in the "value" property. Unfortunately, that value was not available to t-columns, so this adds the "value" to as a slot prop so t-colum

Screen Shot 2022-08-16 at 4 49 02 PM (2)
ns can reference it.

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -766,38 +766,25 @@ export default {
       return index >= this.startIndex && index < this.endIndex;
     },
     getRawCellValue(row, column) {
-      let cellValue = _.get(
+      const cellValue = _.get(
         row.originalRow,
         column.property,
         column.default_value
       );
-      if (cellValue && typeof cellValue === "object") {
-        if (cellValue.value){
-            cellValue = cellValue.value;
-        } 
-        else cellValue = "";
-      }
-      return cellValue;
+      return cellValue && typeof cellValue === "object" ? cellValue.value : '';
     },
     getCellValue(row, column) {
-      const cellValue = this.getRenderedCellValue(row, column);
       return column.format
-        ? column.format(cellValue, row.originalRow)
-        : cellValue;
+        ? column.format(this.getRenderedCellValue(row, column), row.originalRow)
+        : this.getRenderedCellValue(row, column);
     },
     getRenderedCellValue(row, column){
-        let cellValue = _.get(
+        const cellValue = _.get(
         row.originalRow,
         column.property,
         column.default_value
       );
-      if (cellValue && typeof cellValue === "object") {
-        if (cellValue.rendered){
-            cellValue = cellValue.rendered;
-        } 
-        else cellValue = "";
-      }
-      return cellValue;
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -146,6 +146,7 @@
                 :row="row.originalRow"
                 :value="getRawCellValue(row, column)"
                 :rendered_value="getCellValue(row, column)"
+                :rendered="getRenderedCellValue(row, column)"
```
dbroudy, line 149: If I'm following it, `rendered_value` / `getCellValue` is really client rendered, or if not present, server rendered. And `rendered` / `getRenderedCellValue` is purely server rendered. Is there a case where you'd want to use a server rendered value when a client format is specified? If not, would two of these suffice and can naming be aligned? It would be great to get rid of the confusingly named `rendered_value`, since they are almost opposites in this context.

visualizations/frontend/other/TTable.vue
```diff
@@ -771,17 +772,33 @@ export default {
...
+        if (cellValue.value){
+            cellValue = cellValue.value;
+        } 
         else cellValue = "";
       }
       return cellValue;
     },
     getCellValue(row, column) {
-      const cellValue = this.getRawCellValue(row, column);
+      const cellValue = this.getRenderedCellValue(row, column);
```
dbroudy, line 783: Only used of `column.format` is false, maybe move inside the ternary operator?

visualizations/frontend/other/TTable.vue
```diff
@@ -771,17 +772,33 @@ export default {
...
+        let cellValue = _.get(
+        row.originalRow,
+        column.property,
+        column.default_value
+      );
+      if (cellValue && typeof cellValue === "object") {
+        if (cellValue.rendered){
+            cellValue = cellValue.rendered;
+        } 
+        else cellValue = "";
```
dbroudy, line 798: Should this fallback to `cellValue.value`? I don't think there's ever a case where `rendered` and `value` aren't both present, so this whole method might be more complex than necessary.

dbroudy (approved):
This will work, so I'm approving, but I put some comments in that I believe would clean it up.

visualizations/frontend/other/TTable.vue
msorber: You are correct! I think removing rendered_value is the best course of action given that there is not currently a way to pass a format function to the column configuration anyway, and probably won't be for a while. Unfortunately the insights table is referencing rendered_value in several places so I can't remove it safely just yet, but I'll talk to Kennan and get that updated and then remove it.

visualizations/frontend/other/TTable.vue
msorber: Updated

visualizations/frontend/other/TTable.vue
msorber: Awesome, if both rendered and value are always present then this can absolutely be simplified. updated. 👍

PR #116, opened 2022-08-17
feat: Clear all filters button

sareg0:
Jira Ticket: https://snyksec.atlassian.net/browse/DP-312

### Testing locally 
* run the `topcoat-builder-fe` locally and change the config to point at the branch `[sr-testing-clear-all-filters](https://github.com/snyk/topcoat-core/pull/493)`. Rebuild and sync modules once you've updated the config

### Testing in Snyk Insights
Change to the branch `sara-test-branch` in https://snyk-insights.topcoatdata.app/a/develop 
Go to https://snyk-insights.topcoatdata.app/issues-detail. Make sure some filters are visible in the URL
You should see 'Reset Filters' as an option next to the filter dropdowns (see fig. 1.1)

https://github.com/snyk/snyk-insights-topcoat/tree/sara-test-branch

### Testing in app.snyk.io
**WILL NOT WORK YET**  
Waiting until next week for PR to be merged: https://github.com/snyk/registry/pull/28656**

#### Rollout plan/Manual testing plans
If all PRs are approved 
1. merge the topcoat-core PR and this PR, 
2. test that things work as expected in https://snyk-insights.topcoatdata.app. Do not add to production, test only.
3. If all good merge in Registry PRs
4. Test  https://snyk-insights.topcoatdata.app in Reporting Beta in registry
5. Once satisfied, add the 'Reset filters' button to Issues Detail or all beta users and merge into 'dev' or 'main' in https://github.com/snyk/snyk-insights-topcoat

### Behaviour to test
* Click 'Reset Filters' should reset all filters present in the url params, except for `context[org]` or `context[group]`. 
* After the filters are reset the page automatically reloads visualizations.
* When no filters are present the url stays the same. No visualizations reloads should occur.

### Open Question
* UX: There is a slight delay between when the button is clicked and when the page responds and reloads elements. Should we disable the reset filters button after click? 
* UX: should this reset filters button be hidden when no filters are applied?
* Should 'modifiable columns' url query params be preserved or reset?
* Should `context[page]` url query params be preserved or reset?

### Merge order of related PRs: 
1. https://github.com/snyk/topcoat-core/pull/493
2. This PR
3. https://github.com/snyk/registry/pull/28656

Screenshot 2022-08-23 at 15 54 34
fig 1.1


visualizations/frontend/ResetAll.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <button @click="handleClick">{{ text }}</button>
```
sareg0, line 2: I deliberately didn't style this. 

I feel like we need a 'Link as button' component that was the styling, and we use that inside a component that takes care of handling the click. So I went with a proof of concept. This should be replaced with a nice looking component.

visualizations/frontend/ResetAll.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
sareg0, line 1: I wasn't sure where the right place to store this component was. I put it in `visualizations/frontend/ResetAll.vue`, but would be open to suggestions.

visualizations/frontend/other/TResetFilters.vue
```diff
@@ -0,0 +1,17 @@
...
+
+<script>
+export default {
+  data() {
+    is_filter: true
+  },
+  methods: {
+    handleClick(e) {
+      console.log('handle the cliiiiick!', e)
+      this.unsetAllFilters()
```
sareg0, line 13: If I put this component in 'library' it complains that 'unsetAllFilters' is not a function. My assumption is that the VisualizationMixin is not available and can't be called when it is not in the visualizations folder. 

When the component is in the 'visualizations' folder, it has the HeightWrapper applied, which makes the component very big. `is_filter` seems to be the current hack for avoiding the HeightWrapper. When I use `is_filter`, however, the code throws two errors:

```
TypeError: e is undefined // for the 'e' argument to handleClick
ReferenceError: is_filter is not defined
```

I could do with some help understanding how where this component should sit, or if we need to refactor the code in topcoat-core.

visualizations/frontend/other/TResetFilters.vue
```diff
@@ -0,0 +1,17 @@
...
+</template>
+
+<script>
+export default {
+  data() {
+    is_filter: true
+  },
+  methods: {
+    handleClick(e) {
+      console.log('handle the cliiiiick!', e)
```
sareg0, line 12:
```suggestion
```

.gitignore
```diff
@@ -4,4 +4,5 @@ project/*
 static/*
 dbt_gen/*
 *.yml
-.dccache
\ No newline at end of file
+.dccache
+.DS_Store
```
sareg0, line 8: this always pops up for me in my editor, so I thought it was worth while adding it to the ignore file

library/components/TTextButton.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <button class="
```
sareg0, line 2: BaseButton kept on compiling to `basebutton` in the html which has just a `div` with a fancy name. So, to avoid struggling with it I used `button`.

.gitignore
msorber: .DS_store looks like a [mac os metadata file](https://en.wikipedia.org/wiki/.DS_Store), agreed it should be ignored. :D

msorber (approved):

library/components/TTextButton.vue
msorber: Not worth holding up merging, but did you try basse-button? I’m wondering if the back end isn’t doing something odd when building.

sareg0: I'll give it another whirl and see.

sareg0: It works! Must have just been having a moment when I was moving compnents about. I removed the props and the export default since it only relies on a slot.

PR #117, opened 2022-08-22
Feat/ampli analytics

gau94rav:
This PR adds tracking functions to CSV, PDF and Report Dropdown components.

This PR depends on this [PR](https://github.com/snyk/topcoat-core/pull/490) inside topcoat-core.

### DIFF
--- a/library/components/TNavigationDropdown.vue
+++ b/library/components/TNavigationDropdown.vue
@@ -43,7 +43,7 @@
 					:key='index'
 					:class="navSelected(page.url, 'text-[#1c1c21]', 'text-[#555463]')"
 				>
-					<a :href="page.url" class="flex items-center justify-between w-full">
+					<button @click="openReport(page)" class="flex items-center justify-between w-full">
 						{{ page.title }}
 						<span
 							v-if="selected === page.url"
@@ -52,7 +52,7 @@
 						>
 							<check-icon />
 						</span>
-					</a>
+					</button>
 				</li>
 			</ul>
 		</div>
@@ -94,6 +94,19 @@
 			selected() {
 				return window.location.pathname.substring(1);
 			},
+		},
+		methods: {
+			async openReport(page) {
+				await this.trackReportOpen(page.title);
+				window.location.href = page.url;
+			},
+			async trackReportOpen(reportTitle) {
+				if (window.ampli) {
+					await window.ampli.reportIsSelected({
+						reportTitle,
+					})
+				}
+			}
 		},
 	}
 </script>
--- a/library/components/TNavigationDropdown.vue
+++ b/library/components/TNavigationDropdown.vue
@@ -43,7 +43,7 @@
 					:key='index'
 					:class="navSelected(page.url, 'text-[#1c1c21]', 'text-[#555463]')"
 				>
-					<button @click="openReport(page)" class="flex items-center justify-between w-full">
+					<a :href="page.url" class="flex items-center justify-between w-full">
 						{{ page.title }}
 						<span
 							v-if="selected === page.url"
@@ -52,7 +52,7 @@
 						>
 							<check-icon />
 						</span>
-					</button>
+					</a>
 				</li>
 			</ul>
 		</div>
@@ -94,19 +94,6 @@
 			selected() {
 				return window.location.pathname.substring(1);
 			},
-		},
-		methods: {
-			async openReport(page) {
-				await this.trackReportOpen(page.title);
-				window.location.href = page.url;
-			},
-			async trackReportOpen(reportTitle) {
-				if (window.ampli) {
-					await window.ampli.reportIsSelected({
-						reportTitle,
-					})
-				}
-			}
 		},
 	}
 </script>
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -37,14 +37,8 @@
             downloadPdf() {
                 const url = this.page.url + window.location.search;
                 this.downloadPdfFile(url);
-                this.trackPdfExport();
+                this.trackExport("pdf");
             },
-            trackPdfExport() {
-                window.ampli.reportIsExported({
-                    reportExportType: "pdf",
-                    reportTitle: this.page.title,
-                })
-            }
         }
     } 
 </script>
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -29,7 +29,12 @@
         methods: {
             downloadPdf() {
                 const url = this.page.url + window.location.search;
-                return this.downloadPdfFile(url);
+                this.downloadPdfFile(url);
+                window.ampli.reportDataExport({
+                    reportExportType: "pdf",
+                    reportName: this.page.title,
+                    reportLayerName: "null",
+                })
             }
         }
     } 
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -30,10 +30,12 @@
             downloadPdf() {
                 const url = this.page.url + window.location.search;
                 this.downloadPdfFile(url);
-                window.ampli.reportDataExport({
+                this.trackPdfExport();
+            },
+            trackPdfExport() {
+                window.ampli.reportIsExported({
                     reportExportType: "pdf",
-                    reportName: this.page.title,
-                    reportLayerName: "null",
+                    reportTitle: this.page.title,
                 })
             }
         }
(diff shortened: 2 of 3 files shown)


library/components/TPdf.vue
```diff
@@ -29,7 +29,12 @@
         methods: {
             downloadPdf() {
                 const url = this.page.url + window.location.search;
-                return this.downloadPdfFile(url);
+                this.downloadPdfFile(url);
+                window.ampli.reportDataExport({
+                    reportExportType: "pdf",
+                    reportName: this.page.title,
+                    reportLayerName: "null",
+                })
```
sareg0, line 37: how would you feel about moving the event into the mixin in topcoat core? I feel like it could be good for keeping all the implementation details of the tracking library inside topcoat-core, and try to keep the components free of information about the tracking library.

gau94rav: We will still have some events in public though, like report dropdown one.

library/components/TNavigationDropdown.vue
```diff
@@ -95,6 +95,19 @@
...
+				window.location.href = page.url;
+			},
+			async trackReportOpen(reportTitle) {
+				if (window.ampli) {
+					await window.ampli.reportIsSelected({
+						reportTitle,
+					})
+				}
+			}
+		},
```
sareg0, line 110: is there any risk that the ampli function could fail or not complete? Do we need to await it?

library/components/TPdf.vue
sareg0: This is true, but we could emit an event and deal with it in the mixin, if I'm not mistaken?

gau94rav: Yes that might work though we will have to include tracking functions in all mixins depending on the component (vis_mixin, lib_mixin etc), though does make sense to include them all in core.

library/components/TNavigationDropdown.vue
```diff
@@ -43,7 +43,7 @@
 					:key='index'
 					:class="navSelected(page.url, 'text-[#1c1c21]', 'text-[#555463]')"
 				>
-					<a :href="page.url" class="flex items-center justify-between w-full">
+					<button @click="openReport(page)" class="flex items-center justify-between w-full">
```
sareg0, line 46: If I understand correctly, you would like to know when a user opens another report, using this element?
Is that something we can collect using the page visit we already gather when a page is opened?

sareg0: Alternatively, if you can still fire the tracking event and also leave this as a link that would be grand. I am not familiar with triggering an onclick and also a navigation but there are some guidelines here: 

https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR35

I tried it out, but I honestly am not sure how this will behave with our iframe situation: https://codepen.io/sareg0/pen/BaxBEPN?editors=1001

library/components/TNavigationDropdown.vue
sareg0: Hmm... Indeed that is a problem with having the events in topcoat-public and not topcoat-core; things could get out of date. Do we have any method of keeping things up to date that is anything other than manual?

sareg0: Given that it is a page navigation, can we collect the page visit data we already gather when a page is opened and use that? I recall Jeff saying that we have all that information in Amplitude.

I also left a question about this event in the event tracking plan, to clarify what the question is we are answering with this event: https://docs.google.com/presentation/d/1Bv3esoSk-MVA_csGBtS8sb11Z46dwf2zML3O7_y7n_8/edit#slide=id.g13e10e6245e_0_37

gau94rav: hi @sareg0 left the comment in docs.

library/components/TNavigationDropdown.vue
gau94rav: The problem is, this event is specially made for this dropdown and I doubt we can track it as anchor tag, because while we can detect @click, we cannot detect open in new tab.

And maybe I am wrong in this but these this dropdown is more like a selector that is switching reports (from an user's perspective) instead of navigation to a different page.

library/components/TPdf.vue
gau94rav: @sareg0 I realised moving to mixins might not be good, because we might get more components that can fire these events and some of them will be inside vis_mixin while others will be in lib_mixin which might be hard to manage later on, what do you think?

Imagine another export component that is inside visualizations instead of library.

sareg0 (approved):

library/components/TNavigationDropdown.vue
sareg0: Perhaps we need to introduce a new mixin. There may also be a future where we wrap all of our components, and this wrapper is responsible for tracking and what not. 🤷 It's hard to tell where we're going so going with this is fine for now.

sareg0: > And maybe I am wrong in this but these this dropdown is more like a selector that is switching reports (from an user's perspective) instead of navigation to a different page.

Still, anything that changes navigation is most suitable as a link.

library/components/TNavigationDropdown.vue
egcastro: If the idea behind this is tracking when a report is opened, we can use this maybe: https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/#page-view-tracking

gau94rav: This is more of a dropdown interaction, also I think snyk uses their own page tracking in Ampli and they refused to accept my event on this before.

egcastro: I would like to check this. We should not await for this event, we should check Amplitude docs about events flushing when leaving the page or we should use the `page-view-tracking` instead.

sareg0: ` If the Promise is rejected, the await expression throws the rejected value. ` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await


If I understand correctly how [JS handles a throw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw), the execution may be halted unless we deal with any thrown errors, and the navigation could never occur. 


I see that in registry they have written a wrapper that can help with avoiding a slow network request from blocking navigation for too long: https://github.com/snyk/registry/blob/06944be69a2dd4c91fb8a75425cbcfa5162c4e6a/frontend/src/lib/before-navigation-change.ts


If we want to make this navigation tracking more resilient (and we can't use a page visit) then we might want to look into something like the `before-navigation-change` in registry OR do the error handling here and do a 'finally' where we navigate regardless of the promise resolution.

gau94rav: Instead of navigation change event, catching the error sounds better to me, what do you think?

gau94rav: Also if we use the change event, two problems. It will trigger direct change to browser url too, if I am correct, but this event is just for navigation dropdown on pages.

egcastro: I think that we should not `await` any tracking event. Event tracking agents tends to buffer events and send them in batches when possible.
If we want to ensure events to be sent before a page is re-load or user leaves, we could use something like: https://www.docs.developers.amplitude.com/data/sdks/typescript-browser/migration/#flush-or-onexitpage

sareg0: I think that Amplitude can also be changed to use the Beacon API as the `transport` option: https://developers.amplitude.com/docs/advanced-settings#using-sendbeacon

gau94rav: Well in this case, since we should use the 'page exit' approach using, I think it's best to move this event to inside topcoat-core and fire it on page load inside `page_mixing.js` mounted.

We can also move pdf and csv export events to topcoat-core as @sareg0 suggested before inside library_mixin, another benefit of this is that if all the code is inside one project, we can run `ampli status` for health check as well.

Thoughts?

gau94rav: So instead of tracking which page user tried to open, just track which page was opened, this will also include initial page load as well.

sareg0: Even using the Beacon API as the `transport` method might help avoid most problems: https://developers.amplitude.com/docs/advanced-settings#using-sendbeacon

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

gau94rav: @sareg0  @egcastro moved tracking functions to topcoat-core mixins and also added `beacon` transport as well, no longer getting any response, which is good.

Also found out that ampli status is important health check command, so all tracking code should be in core, this was also advised by Sara before and I agree.

PR #118, opened 2022-08-23
Tooltips

gau94rav:
This PR adds new tooltip component for:
1. t-pdf
2. t-url-copy-button

This tooltip can be used to display useful information, the syntax is very simple to use.

```
<t-tooltip v-if="showTooltip" position="left" width="200px">
    Inner content
</t-tooltip>
```

Props:
1. width: Any width value can be used here.
2. position: left, right, bottom, top (default)

Reason for not using the one inside product cl is that it is used as a wrapper for your trigger component, which means it will always show up along with your component that triggers it, even if you have nothing to show, t-pdf for example only shows text when downloading.

### DIFF
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -1,24 +1,32 @@
 <template>
-    <button
-        class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
-        style="height: fit-content;"
-        :disabled="is_loading"
-        @click="downloadPdf"
-    >
-        <div class="flex items-center gap-1">
-            {{ label }}
-            <t-loading-spinner
-                v-if="is_loading"
-                position="relative"
-            />
-        </div>
-    </button>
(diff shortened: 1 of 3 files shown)


library/components/TPdf.vue
```diff
@@ -27,9 +35,15 @@
             },
         },
         methods: {
-            downloadPdf() {
+            async downloadPdf() {
                 const url = this.page.url + window.location.search;
-                return this.downloadPdfFile(url);
+                this.showPdfTooltip = true;
+                await this.downloadPdfFile(url);
```
msorber, line 41: Just out of curiosity, what does the await here accomplish?

msorber (approved):
LGTM

library/components/TPdf.vue
gau94rav: Ahh totally missed it, I needed it before but because it was emitting the state to parent, forgot to remove it, thanks for pointing this out @msorber .

PR #120, opened 2022-08-23
Added prop to use/not use the default url styling and made the additionalUrlParams override the page filters when they include the same parameter

msorber:
To test: 

1. check out the "TURL_testing" branch of snyk-insights
2. click the "Build and Sync modules" icon (cloud between the + and refresh icons)
3. Open: http://localhost:2525/executive_summary?ignored=False&project_name=19377273%257C23023122%257C19174048
4.  Verify that the table at the bottom has 3 rows in the "juice-shop" row, click on the "Open Issues" column.
5.  Verify that the URL in the newly opened tab has the parameter "project_name=19377273", without any additional project ID's
6. Also verify that the styling of the cells in the OPEN ISSUES column is not blue and does not have an underline.

### DIFF
--- a/library/components/TUrl.vue
+++ b/library/components/TUrl.vue
@@ -1,5 +1,5 @@
 <template>
-  <a :href="fullUrl" class="tUrl" :target="openInNewTab ? '_blank' : '_top'" rel="noopener noreferrer">
+  <a :href="fullUrl" :class="{ 'tUrl': includeUrlStyle }" :target="openInNewTab ? '_blank' : '_top'" rel="noopener noreferrer">
     <slot></slot>
   </a>
 </template>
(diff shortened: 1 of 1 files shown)


gau94rav (approved):
LGTM followed instructions in description, seems to be working as expected.

library/components/TUrl.vue
```diff
@@ -25,18 +25,29 @@ export default {
...
+      default: true
     }
   },
   computed: {
     fullUrl() {
-      let allUrlParams = this.additionalUrlParams;
+      let allUrlParams = {...this.additionalUrlParams};
 
       if (this.includeFilterParams ) {
         const { 'context[org]': org, 'context[group]': group, ...restOfFilters } = this.getFiltersState;
```
sareg0, line 39: Non-blocking comment:
oh, interesting!! I added a getter to the global state to get context filters only. We could think about adding a getter to return all non context filters as well so that we can a centralized list of context filters we shouldn't touch: https://github.com/snyk/topcoat-core/blob/b0e61e137925e8427b3df162451d94823c688a25/frontend/src/store/modules/page.js#L85

Thought you might find it interesting, as we proceed adding components that reference context filters.

sareg0 (approved):

library/components/TUrl.vue
msorber: That is a fantastic suggestion! I added a [Jira ticket ](https://snyksec.atlassian.net/browse/DP-326)so we don't forget. Wish that wouldn't be a whole seperate PR or I'd do it as part of this, but such is life.

PR #121, opened 2022-08-25
refactor tooltip code

gau94rav:
Fixes:
1. Current tooltip was having trouble placing it self near it's trigger and needed a better approach
2. is-clickable prop can be used to make tooltip appear on click
3. outside click detection function is used to close tooltips
4. t-pdf and t-url-copy-button are now using this new verison as well.

How to test:
1. Switch your config.yml `revision` to this branch.
2. Click Build button inside IDE.
3. Clicking on pdf download will open the new tooltip.
4. Hovering on t-url-copy-button will open the tooltip.

New syntax:

```
<t-tooltip width="200px" position="top" :is-clickable="false">
   <button slot="trigger">button</button>
   This is the tooltip text, now also supports anchor tags.
</t-tooltip>
```

### DIFF
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -1,33 +1,30 @@
 <template>
-    <div class="relative" @mouseover="handleTooltip" @mouseleave="showPdfTooltip = false">
+    <div class="relative">
         <!-- PDF tooltip -->
-        <t-tooltip v-if="showPdfTooltip" position="left" width="400px">
+        <t-tooltip position="left" width="400px" is-clickable>
+            <button
+                class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
+                style="height: fit-content;"
+                :disabled="is_loading"
+                @click="downloadPdf"
+                :class="is_loading && 'opacity-60'"
+                slot="trigger"
+            >
+                <div class="flex items-center gap-1">
+                    {{ is_loading ? "Generating PDF" : label }}
+                    <t-loading-spinner
+                        v-if="is_loading"
+                        position="relative"
+                    />
+                </div>
+            </button>
             Your PDF is downloading. It may take up to 30 seconds to complete. Remain on the this page for the download to complete.
         </t-tooltip>
-
-        <button
-            class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
-            style="height: fit-content;"
-            :disabled="is_loading"
-            @click="downloadPdf"
-            :class="is_loading && 'opacity-60'"
-        >
-            <div class="flex items-center gap-1">
-                {{ is_loading ? "Generating PDF" : label }}
-                <t-loading-spinner
-                    v-if="is_loading"
-                    position="relative"
-                />
-            </div>
-        </button>
     </div>
 </template>
 
 <script>
     export default {
-        data: () => ({
-            showPdfTooltip: false,
-        }),
         props: {
             label: {
                 type: String,
@@ -37,14 +34,8 @@
         methods: {
             downloadPdf() {
                 const url = this.page.url + window.location.search;
-                this.showPdfTooltip = true;
                 this.downloadPdfFile(url);
             },
-            handleTooltip() {
-                if (this.is_loading) {
-                    this.showPdfTooltip = true;
-                }
-            }
         }
     } 
 </script>
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -18,7 +18,9 @@
                     />
                 </div>
             </button>
-            Your PDF is downloading. It may take up to 30 seconds to complete. Remain on the this page for the download to complete.
+            <div v-if="is_loading">
+                Your PDF is downloading. It may take up to 30 seconds to complete. Remain on the this page for the download to complete.
+            </div>
         </t-tooltip>
     </div>
 </template>
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -54,7 +54,7 @@ export default {
 			// Calculate and place both tooltip and arrow.
 			const triggerElement = this.$refs.trigger;
 			const tooltipElement = this.$refs.tooltip;
-			const tooltipPositions = {};
+
 			const positions = { arrow: {}, tooltip: {} };
 			if (triggerElement && tooltipElement) {
 				const position = triggerElement.getBoundingClientRect();
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -41,7 +41,6 @@ export default {
 	data: () => ({
 		positions: {},
 		showTooltip: false,
-		isForcedOpen: false
 	}),
 	mounted() {
 		onClickOutside(this.$refs.tooltipContainer, () => {
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -1,16 +1,22 @@
 <template>
-	<div @mouseover="handleMouseOver" @click="handleClick" ref="tooltipContainer">
+	<div
+        @mouseover="handleMouseOver"
+        @mouseleave="handleMouseLeave"
+        @click="handleClick"
+        ref="tooltipContainer"
+    >
 		<div ref="trigger">
 			<slot name="trigger">
 				<help-circle-outline-icon :size="12" />
 			</slot>
 		</div>
+
+        <!-- $slots.default is used in cases when the trigger is using v-if on content's div -->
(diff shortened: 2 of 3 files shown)


library/components/TTooltip.vue
```diff
@@ -1,64 +1,122 @@
...
+				this.show();
+			}
+		},
+		show() {
+			this.showTooltip = true;
+
+			// Glitchy behaviour without nextTick()
+			this.$nextTick(() => {
+				this.placeTooltip();
+			});
```
gau94rav, line 115: Unfortunately unable to avoid this for now.

sareg0: Please explain further?

library/components/TTooltip.vue
```diff
@@ -1,64 +1,122 @@
...
-		tooltipPositions: {}, // This object should be empty by default to only apply related properties (top, left...)
-        arrowPositions: {
-            top: "left-0 right-0 -bottom-[7px]",
-            bottom: "left-0 right-0 -top-[7px]",
-            left: "top-2 -right-[7px]",
-            right: "top-2 -left-[7px]",
-        }
+		positions: {},
+		showTooltip: false,
+		isForcedOpen: false
```
sareg0, line 44:
```suggestion
```
I didn't see if being used

library/components/TTooltip.vue
```diff
@@ -1,64 +1,122 @@
...
+					positions.arrow.top = position.top + "px";
+					positions.arrow.left = position.left - 25 + "px";
+				} else {
+					positions.tooltip.top =
+						position.top - tooltipElement.clientHeight / 2 + "px";
+					positions.tooltip.left =
+						position.left + triggerElement.clientWidth + 15 + "px";
+					positions.arrow.top = position.top + "px";
+					positions.arrow.left =
+						position.left + triggerElement.clientWidth + 10 + "px";
```
sareg0, line 91: I think the diff seems bigger than it is becaues the indentation is going from 2 spaces to 4 spaces. Do you think you could run the formatter to put the indentation back to 2 spaces?

library/components/TPdf.vue
```diff
@@ -1,33 +1,30 @@
 <template>
-    <div class="relative" @mouseover="handleTooltip" @mouseleave="showPdfTooltip = false">
+    <div class="relative">
         <!-- PDF tooltip -->
-        <t-tooltip v-if="showPdfTooltip" position="left" width="400px">
+        <t-tooltip position="left" width="400px" is-clickable>
```
sareg0, line 4: after clicking the button to download the pdf, the tooltip never comes back.
https://user-images.githubusercontent.com/1307818/186701000-7e7f6bf8-1cab-4a2e-946a-c09c3e061033.mov

library/components/TTooltip.vue
```diff
@@ -1,64 +1,122 @@
...
+			<slot name="trigger">
+				<help-circle-outline-icon :size="12" />
+			</slot>
+		</div>
+		<div
+			class="p-4 bg-[#1C1C21] text-white w-max h-max rounded fixed z-50"
+			:style="{ width, ...positions.tooltip }"
+			ref="tooltip"
+			v-show="showTooltip"
+			@mouseleave="handleMouseLeave"
```
sareg0, line 13: I think that all the events should be on the trigger. As that is what is initiating the tooltip

library/components/TPdf.vue
gau94rav: In the case of t-pdf I am using the `is-clickable` prop so it only shows up on click, but this will confusing isn't it, yes should be handled better I think.

library/components/TTooltip.vue
```diff
@@ -1,64 +1,119 @@
...
+		>
+			<slot></slot>
```
sareg0, line 15: I think the default slot should be for the trigger element. And the message displayed in the tooltip should be an enforced prop. This would avoid being able to use the tooltip without a message: 

ProductCL has a good example: https://github.com/PatchUI/ProductCL/blob/main/src/components/BaseTooltip/BaseTooltip.vue

library/components/TPdf.vue
sareg0: I am used to tooltips behaving like this: https://dequeuniversity.com/library/aria/tooltip

I think hovers and focus are enough for our uses cases, no?

gau94rav: @sareg0  I see, I like these examples, if hovered on a button, then hide on `mouseleave` but if focused, keep it until user clicks away.

library/components/TTooltip.vue
gau94rav: The problem with using prop is that, it is only limited to text, but according to one of the slide I saw inside figma, our tooltips can also have clickable links, this was a big reason why I went with custom tooltips.

sareg0: Ah, ok. 

TIL: clickable content can be ok in a tooltip: https://www.nngroup.com/articles/tooltip-guidelines/ (section 3)

library/components/TTooltip.vue
```diff
@@ -1,64 +1,110 @@
...
+			this.show();
+		},
+		handleMouseLeave() {
+			if (this.isFocused) return;
+			this.hide();
+		},
+		handleClick() {
+            this.isFocused = true;
+            this.show();
+		},
```
sareg0, line 96: It might be hard to keep `isFocused` up to date with the focused state of an element.

How do you feel about changing tooltip to provide a prop (`open`), to force the tooltip open. If I recall rightly you had something like this in an earlier implementation. The implementation of ProductCL is a good example: https://github.com/PatchUI/ProductCL/blob/main/src/components/BaseTooltip/BaseTooltip.vue#L115

Tooltip can still be responsible for its mouseover/mouseleave and potentially a click. However, a trigger element can tell the tooltip to display when its native html events are triggered

for example, focusing a button:
```
<tooltip :open="showTooltip">
  <button @focus="showTooltip = true" @blur="showTooltip=false"></button>
</tooltip>
```

That way, you can rely on the `open` prop inside the tooltip to decide whether to `return` on the mouse events and click events, and leverage native DOM events to open and close the tooltip when you have them available to you.

library/components/TUrlCopyButton.vue
```diff
@@ -1,26 +1,25 @@
...
+        <t-tooltip position="bottom" width="130px">
+            <button
+                class="h-[30px] w-[30px] rounded-[4px] flex items-center justify-center border border-[#C3C2CB] p-1 transition-colors"
+                @click="copy"
+                :class="{ 'bg-[#B4E4D9]': copied }"
+                slot="trigger"
+            >
+                <check-icon :size="18" v-if="copied" />
+                <link-icon :size="18" v-else />
+            </button>
```
sareg0, line 13: NON-BLOCKING SUGGESTION:
### Accessibility improvement

This button is an icon only, but has text that we could integrate. We could improve its accessibility by moving the text into the button and hiding it

```suggestion
            <button
                class="h-[30px] w-[30px] rounded-[4px] flex items-center justify-center border border-[#C3C2CB] p-1 transition-colors"
                @click="copy"
                :class="{ 'bg-[#B4E4D9]': copied }"
                slot="trigger"
            >
                <check-icon :size="18" v-if="copied" />
                <link-icon :size="18" v-else />
                <span class="visually-hidden">{{ copied ? "URL Copied" : "Copy URL" }}</span>
            </button>
```


```
/* 
 * Utility class to hide content visually while keeping it screen reader-accessible.
 * Source: https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html 
 */

.visually-hidden:not(:focus):not(:active) {
  clip: rect(0 0 0 0); 
  clip-path: inset(100%); 
  height: 1px; 
  overflow: hidden; 
  position: absolute; 
  white-space: nowrap; 
  width: 1px; 
}
```

This technique comes from this article: https://www.sarasoueidan.com/blog/accessible-icon-buttons/#technique-%231%3A-accessible-visually-hidden-text

There are several other techniques, but this seems like the most straight forward one for the moment. We could add the `visually-hidden` css utility to our global stylesheets for future use.

sareg0 (approved):
Feel free to move forward with this whenever you're happy with it. 
I left some suggestions, but I will leave it to your discretion to implement them or not as it makes sense for your use cases.

library/components/TTooltip.vue
gau94rav: This makes sense but the problem is the page files like `issues_detail.html` do not support javascript at all, this might not work for them and will make this feature exclusive to library or visualization components.

Also the reason I removed it was also because it might require more syntax for non-tech users, like Kennan or Ezra might not like this haha!

Though I still want to keep this in mind, maybe we can merge them in future.

PR #122, opened 2022-08-25
various big number fixes

gau94rav:
New big number was developed in rush with many ideas that make it too complex, this PR aims to clean it a bit and also fix several issues.

It also solves this [jira](https://snyksec.atlassian.net/jira/software/c/projects/DP/boards/443?modal=detail&selectedIssue=DP-325&assignee=62261dde94f7e200690091ae) ticket as well.

The tooltip part for big number in this PR depends on this [PR](https://github.com/topcoat-data/topcoat-public/pull/121).

### DIFF
--- a/visualizations/frontend/other/TBigNumber.vue
+++ b/visualizations/frontend/other/TBigNumber.vue
@@ -1,174 +1,212 @@
 <template>
-    <div
-        class="w-full lg:p-5 md:p-3 p-1 rounded-[2px] break-all"
-        :style="{ background: bgColor, color: textColor }"
-        :class="hasUnderline && 'border-b-2 border-[#21214c]'"
-    >
-        <div class="flex items-center gap-2">
-            <div class="hidden opacity-50 md:block" v-if="hasIcon">
-                <slot name="icon"></slot>
-            </div>
-            <span class="text-xs font-semibold leading-4 tracking-widest">
-                {{ label || '--' }}
-            </span>
-            <base-tooltip v-if="tooltipText" :description="tooltipText" />
-        </div>
-        <div :class="hasTallSize ? 'mt-6' : 'mt-3'">
-            <span class="font-normal leading-8" :class="fontSizes[bigNumberSize]">
-                <div class="flex items-center gap-2">
-                    <span  :style="{ color: numberTextColor }">{{ value || '--' }}</span>
-                    {{ valueText }}
-                    <div v-if="previous">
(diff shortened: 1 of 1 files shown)


sareg0:
Are there any core changes we should be aware of? 
Can you give some testing insctructions?

gau94rav:
> Are there any core changes we should be aware of? Can you give some testing instructions?



> Are there any core changes we should be aware of? Can you give some testing instructions?

Hi @sareg0 the easiest way to test this is by switching to this branch and using this syntax:

`<t-big-number :number-value="29" :number-previous="40" label="Big Num" />`

This component can easily be used without any layer.

But if you want to use with a layer you can use this simple query for test.

`SELECT 28 as value, 40 as previous`

library/components/TTooltip.vue
```diff
@@ -38,14 +38,19 @@ export default {
 			type: String,
 			default: "top"
 		},
+		isVisible: { // To control state from a component.
+			type: Boolean,
+			default: false,
+		}
```
gau94rav, line 44: @sareg0 You were right about controlling it from component, but at the same time we need internal state as well for pages 😅, I am thinking about having both, I have seen this approach in other CLs as well, let me know if you have thoughts on it.

gau94rav:
We need a linting tool for public repo,  the code is always badly linted after pushed.

library/components/TTooltip.vue
sareg0: suggestion: split the big number and tooltip fixes

visualizations/frontend/other/TBigNumber.vue
```diff
@@ -1,174 +1,212 @@
...
+					<slot name="tooltip"></slot>
+				</t-tooltip>
+			</div>
+		</div>
+
+		<div class="mt-3">
+			<span class="font-normal leading-8" :class="fontSizes[bigNumberSize]">
+				<div class="flex items-center gap-2">
+					<t-tooltip position="right">
+						<span slot="trigger" :style="{ color: numberTextColor }">
```
gau94rav, line 27: This was a old already merged code of Meg, re-added it to new syntax inside tooltip.

library/components/TTooltip.vue
gau94rav: hi @sareg0 splitted the tooltip [PR](https://github.com/topcoat-data/topcoat-public/pull/134)

sareg0 (approved):
Unfortunately I can't really see what's changing very well. I trust your judgement.

PR #123, opened 2022-08-25
Added includeContextParam prop to TURL

msorber:
To test:

- check out the executive_summary branch of snyk-insights. 
- Open the [executive summary](http://localhost:2525/executive_summary?context%5Borg%5D=4a3d29ab-6612-481b-83f2-aea6cf421ea5&ignored=False&exec_summary_start_date=2022-05-01&exec_summary_end_date=2022-05-31)
- click the "Total Identified" big number
- verify that the url includes the param "context%5Borg%5D=4a3d29ab-6612-481b-83f2-aea6cf421ea5"

### DIFF
--- a/library/components/TUrl.vue
+++ b/library/components/TUrl.vue
@@ -28,16 +28,27 @@ export default {
     },
     includeUrlStyle: {
       type: Boolean,
-      default: true
+      default: true,
+    },
+    includeContextParam:{
+      type: Boolean,
+      default: false
     }
   },
   computed: {
     fullUrl() {
-      let allUrlParams = {...this.additionalUrlParams};
+      let allUrlParams = {};
+      const { 'context[org]': org, 'context[group]': group, ...restOfFilters } = this.getFiltersState;
+
+      if(this.includeContextParam){
+        if(org){
+          allUrlParams['context[org]'] = org
+        }else if(group){
+          allUrlParams['context[group]'] = group
+        }
+      }
 
       if (this.includeFilterParams ) {
-        const { 'context[org]': org, 'context[group]': group, ...restOfFilters } = this.getFiltersState;
-
         // Sometimes the additional URL params will include the same filter as one of the filters already in the URL
         // So the order of the url params below is important, any duplicated keys will end up being the last one added. 
         // 
@@ -46,8 +57,9 @@ export default {
         // url should go to the page with the issues-details filtered on the row's project, not the page where
         // all of the projects selected in the filter are present.
         allUrlParams = {
+          ...allUrlParams,
           ...restOfFilters,
-          ...allUrlParams
+          ...this.additionalUrlParams,
         };
       }
 


library/components/TUrl.vue
```diff
@@ -28,16 +28,27 @@ export default {
     },
     includeUrlStyle: {
       type: Boolean,
-      default: true
+      default: true,
+    },
+    includeContextParam:{
+      type: Boolean,
+      default: false
```
sareg0, line 35: why do we need to give the user a choice? context params are an internal concept, so I wouldn't think users would want to know.

sareg0 (approved):
if you need to merge, it's fine.

library/components/TUrl.vue
msorber: Excellent point, started a conversation in slack. Going to mark this a draft while I think up a better solution.

library/components/TUrl.vue
```diff
@@ -28,16 +28,27 @@ export default {
...
     }
   },
   computed: {
     fullUrl() {
-      let allUrlParams = {...this.additionalUrlParams};
+      let allUrlParams = {};
+      const { 'context[org]': org, 'context[group]': group, ...restOfFilters } = this.getFiltersState;
```
sareg0, line 41: I think that sometimes we there is a `context[page]` in the url. This destructuring would not catch that and would mean `context[page]=issue_detail` (or whatever page name) could be in the generated URL. Would that be a problem?

library/components/TUrl.vue
sareg0: Okay, to respond to your thoughts in Slack in the context of this PR I thought it would be good to leave comment 
https://snyk.slack.com/archives/C0372B66DCM/p1661452777357679

If we DO need the ability to decide whether to include context params, please feel free to keep it. If the prop default is 'false' then users will be relatively unaffected for now. However, I question how much a data analyst cares to control or know about 'context' params. If we plan to use it for every URL then every TURL will have `includeContextParam=true` and we won't be able to tell if users really want it, or just use it to make sure things work correctly. It will propogate because people will copy and paste since it's not immediately clear what the consequence of not adding `includeContextParam` or not means unless a user reads the code 🤷‍♀️  

I think that if you wanted to split the logic and the UI parts of the components you could do so on these lines. It's only an idea, and could create more overhead until the actual need arises.
1. TAnchor - anchor tag, takes a prop for href and several props that map to the html attributes. Basically what TUrl was to begin with. TAnchor is responsible for the way it looks, and any native html things an anchor does. 
4. TLink - this is a smarter component and is what the TUrl component has become. It can use TAnchor and pass it the url once it has all been calculated. This URL could be calculated elsewhere, such as a mixin. Or at the very least the params that will be needed 

If this url calculation is only relevant for TUrl, then that's fine, leave it. If it starts to spread, then maybe we need to put it somewhere re-usable. 

The future is hard to foresee. For right now, it's fine as it is.

library/components/TUrl.vue
msorber: Nope! In this case the page is changing anyway. :D

PR #124, opened 2022-08-25
Added TrendArrow

msorber:
To test, update config.yml to point to this branch and add the following to a page:
`      

        up:<TTrendArrow direction="up" :size="80" />

        neutral:<TTrendArrow direction="neutral" :size="70" />

        down:<TTrendArrow direction="down" :size="60" />

        1:<TTrendArrow :direction="1" :size="50" />

        0:<TTrendArrow :direction="0" :size="40" />

        -1:<TTrendArrow :direction="-1" :size="30" />

        error: <TTrendArrow direction="dksfjdlskfjl" size="30" />
`


sareg0:
Can you give some indication of how it's going to be used? 
What sort of numbers are we going to pass to the component?

msorber:
Of course! :D 

[From the executive summary](https://github.com/snyk/snyk-insights-topcoat/blob/executive-summary/library/CurrentAndPast.vue):
`<TrendArrow :firstNumber="current" :secondNumber="past" class="trendIcon" />`

In the above case the 2 numbers are the current open/new/resolved issues and the past open/new/resolved issues. But this component wouldn't be limited to those- could be a current and future number, expected vs. actual etc.

library/components/TrendArrow.vue
```diff
@@ -0,0 +1,39 @@
...
+    <div class="trendArrow">
+        <arrow-up-icon v-if="difference > 0" :size="iconSize" />
+        <minus-icon v-else-if="difference === 0" :size="iconSize" />
+        <arrow-down-icon v-else :size="iconSize" />
+    </div>
+</template>
+
+<script>
+export default {
+  name: "TrendArrow",
```
sareg0, line 11: Are we appending `T` to all topcoat-public components?

library/components/TrendArrow.vue
```diff
@@ -0,0 +1,39 @@
...
+  props: {
+    firstNumber: {
+      type: Number,
+      required: true,
+    },
+    secondNumber: {
+      type: Number,
+      required: true,
+    },
```
sareg0, line 20: current and past seems descriptive 🤷‍♀️

library/components/TrendArrow.vue
```diff
@@ -0,0 +1,39 @@
...
+    iconSize:{
+      type: String,
+      default: '20'
```
sareg0, line 23: They should be a number no?
otherwise someone could pass in any string and it would get used

library/components/TrendArrow.vue
msorber: ...yes, yes we are. :joy: Will fix!

library/components/TrendArrow.vue
msorber: Agreed, and I _hate_ "firstNumber" and "secondNumber", but "current" and "past" would be confusing if you wanted to use it to show something like "did we get more, the same, or fewer story points fixed this sprint than expected?" On that note "Trend" is probably not the right name either, I just can't come up with anything better. Shake fist @ brain.

sareg0: I actually have another suggestion, but I'll finish writing it. You can choose

library/components/TrendArrow.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+    <div class="trendArrow">
+        <arrow-up-icon v-if="difference > 0" :size="iconSize" />
+        <minus-icon v-else-if="difference === 0" :size="iconSize" />
+        <arrow-down-icon v-else :size="iconSize" />
```
sareg0, line 5: If you wanted you could load the 'smarts' about the first and last number and calculating the difference into CurrentAndPast. This component could take a prop like

```
    type: {
      validator: function (value) {
        // The value must match one of these strings
        return ['up', 'down', 'neutral'].includes(value)
      }
```

thise would avoid duplicating the 'first' and 'second' number that the  `CurrentAndPast` has and we could expand it relatively easily... As well as use it without numbers... Feel free to go with the direction you like though.

sareg0 (approved):
As we move forward with the executive summary I hope we get feedbak on the upward/downard trends. I had a discussion with design eng about labels on the ProductCL icons: https://snyk.slack.com/archives/C02JDRWT06P/p1661519886277659

Their meaning can feel ambiguous. However, we can solve this over time as we get feedback from users. I wanted to raise awareness since you're in the middle of it right now and I thought it might be interesting for you.

Although we get some labels that are taking a default of the icon name, they are not great for this context. Below the aria-label is 'Arrow up icon' for example:
Screenshot 2022-08-26 at 18 33 20
🤷‍♀️ 

It's fine to go ahead with this when you are satisfied with the look. 

We could provide an aria-label now, or we could look at the CurrentandPast element as a whole (my peference) to determine what the best way forward might be for improving the accessibility, at a later date.

library/components/TrendArrow.vue
msorber: Reducing the input to a single value is a fantastic idea! I think having it stay a number makes sense from an ease of use standpoint, but there is no reason a user can't do the difference calculation themselves which would neatly sidestep the bad prop names issue. :D

library/components/TrendArrow.vue
msorber: Yep! The format used elsewhere was string so I just assumed it was a string (maybe the icons accept input like "small" "medium" and "large"?) but numbers work and make more sense.

sareg0 (approved):

library/components/TrendArrow.vue
msorber: Agreed with the above. Additionally there is a non-trivial concern that users will see up/neutral/down as a prop and assume that those are the _only_ valid values.

PR #125, opened 2022-08-25
added no_data slot to TTable

msorber:
To test check out the add_issues_details_no_data branch, in the issues details page select filters until there are no matching issues and instead of "no data" you should see table contents that match [this](https://www.figma.com/file/IaywW74sLSgYJSYHjOC0ZL/Reporting-Beta-2022?node-id=2882%3A47764).

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -92,8 +92,10 @@
         "
         class="spanAllColumns center_cell"
       >
-        <div><i class="i-fa-solid i-fa-inbox"></i></div>
-        <div>{{ noDataMessage }}</div>
+        <slot name="no_data" >
+          <div><i class="i-fa-solid i-fa-inbox"></i></div>
+          <div>{{ noDataMessage }}</div>
+        </slot>
       </div>
 
       <!-- Data Rows -->


visualizations/frontend/other/TTable.vue
```diff
@@ -92,8 +92,10 @@
         "
         class="spanAllColumns center_cell"
       >
-        <div><i class="i-fa-solid i-fa-inbox"></i></div>
-        <div>{{ noDataMessage }}</div>
+        <slot name="no_data" >
+          <div><i class="i-fa-solid i-fa-inbox"></i></div>
```
sareg0, line 96: I don't see any icons in the design. Is this needed?

In addition, I think we might be moving away from fa icons and using more productcl ones, no?

sareg0 (commented):
It does not match the design

Screenshot 2022-08-26 at 13 17 00

### Testing setup 

point to this branch in the builder config on https://snyk-insights.topcoatdata.app/a/develop
```
modules:
      revision: add_no_data_slot_to_table
```

point to following git branch https://github.com/snyk/snyk-insights-topcoat/tree/add_issues_details_no_data

visualizations/frontend/other/TTable.vue
msorber: The icon is from the [pre-snyk table design](https://docs.google.com/spreadsheets/d/1854rrORMna_BKL7u73I140G9rtrivtcKVNi6EvMSXjs/edit#gid=0). 

Yeah I think we are moving away from fa icons, should probably be its own pitch to go through all the topcoat-public components and replace icons/remove fa dependency.

msorber:
That is odd, I see this with the above setup:

Screen Shot 2022-08-26 at 10 02 26 AM

sareg0:
Okay, I had to rebuild the modules 

Screenshot 2022-08-26 at 17-57-31 Issues Detail

visualizations/frontend/other/TTable.vue
sareg0: I don't see the icons, do you?

sareg0 (approved):
Do I understand correctly that you are providing a slot for it, because you want different messages per usage of the table?

https://github.com/snyk/snyk-insights-topcoat/pull/17

msorber:
> 

Yep! And the props are not flexible enough to support the above design, hence the need for a slot.

PR #126, opened 2022-08-26
Default value props for date picker

gau94rav:
Adds new props for date picker.

1. `default-date-preset`: Will apply the preset and dates accordingly, can accept both `key` and `label`.
2. `default-start-date`: Will apply the date as start date, also respects `mode(range, single)` .
3. `default-end-date`: If empty but start date has default, then same is applied to this one to avoid any glitch.

Note:
- If default preset given then start and end date defaults are skipped.
- Using default dates will automatically set preset to custom (if preset default not used).

### DIFF
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -282,7 +282,6 @@ export default {
             const defaultEndDate = this.defaultEndDate || defaultStartDate; // If no end date default, set same as start to avoid error.
 
             if (defaultStartDate && defaultEndDate) {
-                console.log(window.Moment(defaultStartDate).toDate())
                 const date = [
                     window.Moment(defaultStartDate).toDate(),
                     window.Moment(defaultEndDate).toDate()
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -197,12 +197,10 @@ export default {
                 
                 case "mtd":
                     startDate = startDate.startOf('month');
-					endDate = endDate.subtract(1, 'days');
                     break;
                 
                 case "ytd":
 					startDate = startDate.startOf('year');
-					endDate = endDate.subtract(1, 'days');
                     break;
             }
 
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -58,6 +58,18 @@ export default {
             type: String,
             default: " ~ ",
         },
+        defaultStartDate: {
+            type: String,
+            default: "",
+        },
+        defaultEndDate: {
+            type: String,
+            default: "",
+        },
+        defaultDatePreset: {
+            type: String,
+            default: "",
+        }
     },
     computed: {
         mode() {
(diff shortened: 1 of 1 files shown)


sareg0:
Why are you adding this? Was there a request to add a default?

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -233,6 +246,54 @@ export default {
...
+
+            if (urlStartDate || urlEndDate) {
+                return;
+            }
+
+            const defaultStartDate = this.defaultStartDate;
+            const defaultEndDate = this.defaultEndDate || defaultStartDate; // If no end date default, set same as start to avoid error.
+
+            if (defaultStartDate && defaultEndDate) {
+                console.log(window.Moment(defaultStartDate).toDate())
```
sareg0, line 285:
```suggestion
```

gau94rav:
> Was there a request to add a default?

Totally forgot about it 😅, yes this was requested from Kennan.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -233,6 +246,54 @@ export default {
...
+        handleDefaultDates() {
+            const urlStartDate = this.getFilterValue("start_date");
+            const urlEndDate = this.getFilterValue("end_date");
+
+            if (urlStartDate || urlEndDate) {
+                return;
+            }
+
+            const defaultStartDate = this.defaultStartDate;
+            const defaultEndDate = this.defaultEndDate || defaultStartDate; // If no end date default, set same as start to avoid error.
```
sareg0, line 282: If 'defaultStartDate' is not supplied it will be defaulted to an empty (via the props 'default'). This means the `defaultEndDate` will also be an empty string. Is that ok?

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -58,6 +58,18 @@ export default {
...
+            default: "",
+        },
+        defaultEndDate: {
+            type: String,
+            default: "",
+        },
+        defaultDatePreset: {
+            type: String,
+            default: "",
+        }
```
sareg0, line 72: What potential values could user supply for `defaultDatePreset`?

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: Yes, because if both are empty, the whole default logic will be skipped.

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: It should be the value that matches with your provided `dateFormat` prop, if format is `YYYY-MM-DD` then default should be `2022-12-31`

sareg0: so `defaultDatePreset` would be `2022-12-13`?

gau94rav: Oh wait sorry totally missed it, no actually this one is a bit hard without documentation but this should just match the preset keys or labels.

- defaultDatePreset="ytd"

gau94rav: Both will work, so any user can open dropdown, see what presets are available and type it's name, they do not have to know they key at all.

sareg0: Okey dokey. Yeah, we might need to document it... Or perhaps the `defaultDatePresets` could be an enum sometime in the future 🤷‍♀️  Let's see if it evolves

sareg0 (approved):

PR #130, opened 2022-09-07
Range Input throttle

gau94rav:
**Problem**
This component gives users the ability to drag the slider and update the value, which can cause too many request on every change and fills up out tracking data as well.

**Fix**
Added throttling to only update after half second after user stops dragging slider.

### DIFF
--- a/visualizations/frontend/filters/v2/TRangeInput.vue
+++ b/visualizations/frontend/filters/v2/TRangeInput.vue
@@ -76,14 +76,18 @@
         },
         methods: {
             update(e) {
-                if (e.minValue === e.min && e.maxValue == e.max) {
-                    this.unsetFilterValue('min', true);
-                    this.unsetFilterValue('max', true);
-                } else {
-                    this.barMinValue = e.minValue;
-                    this.barMaxValue = e.maxValue;
-                }
-            }
+                if (this.throttleInterval) clearInterval(this.throttleInterval);
+
+                this.throttleInterval = setTimeout(() => {
+                    if (e.minValue === e.min && e.maxValue == e.max) {
+                        this.unsetFilterValue('min', true);
+                        this.unsetFilterValue('max', true);
+                    } else {
+                        this.barMinValue = e.minValue;
+                        this.barMaxValue = e.maxValue;
+                    }      
+                }, 500)
+            },
         },
         mounted() {
             this.fetchLayerData();
--- a/visualizations/frontend/filters/v2/TRangeInput.vue
+++ b/visualizations/frontend/filters/v2/TRangeInput.vue
@@ -78,7 +78,7 @@
         methods: {
             update(e) {
                 if (this.throttleId) {
-                    clearInterval(this.throttleId);
+                    clearTimeout(this.throttleId);
                 }
 
                 this.throttleId = setTimeout(() => {
--- a/visualizations/frontend/filters/v2/TRangeInput.vue
+++ b/visualizations/frontend/filters/v2/TRangeInput.vue
@@ -83,8 +83,8 @@
 
                 this.throttleId = setTimeout(() => {
                     if (e.minValue === e.min && e.maxValue == e.max) {
-                        this.unsetFilterValue('min', true);
-                        this.unsetFilterValue('max', true);
+                        this.unsetFilterValue('min');
+                        this.unsetFilterValue('max');
                     } else {
                         this.barMinValue = e.minValue;
                         this.barMaxValue = e.maxValue;
--- a/visualizations/frontend/filters/v2/TRangeInput.vue
+++ b/visualizations/frontend/filters/v2/TRangeInput.vue
@@ -19,7 +19,6 @@
     export default {
         data: () => ({
             is_filter: true,
-            throttleId: null,
         }),
         props: {
             label: {
@@ -76,21 +75,15 @@
             }
         },
         methods: {
-            update(e) {
-                if (this.throttleId) {
-                    clearTimeout(this.throttleId);
-                }
-
-                this.throttleId = setTimeout(() => {
-                    if (e.minValue === e.min && e.maxValue == e.max) {
-                        this.unsetFilterValue('min');
-                        this.unsetFilterValue('max');
-                    } else {
-                        this.barMinValue = e.minValue;
-                        this.barMaxValue = e.maxValue;
-                    }      
-                }, 250)
-            },
+			update: window._.debounce(function (e) {
+				if (e.minValue === e.min && e.maxValue == e.max) {
+                    this.unsetFilterValue('min');
+                    this.unsetFilterValue('max');
+                } else {
+                    this.barMinValue = e.minValue;
+                    this.barMaxValue = e.maxValue;
+                } 
+            }, 250)
         },
         mounted() {
             this.fetchLayerData();
--- a/visualizations/frontend/filters/v2/TRangeInput.vue
+++ b/visualizations/frontend/filters/v2/TRangeInput.vue
@@ -19,6 +19,7 @@
     export default {
         data: () => ({
             is_filter: true,
+            throttleId: null,
         }),
         props: {
             label: {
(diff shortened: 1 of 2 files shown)


visualizations/frontend/filters/v2/TRangeInput.vue
```diff
@@ -76,14 +76,18 @@
...
+                this.throttleInterval = setTimeout(() => {
+                    if (e.minValue === e.min && e.maxValue == e.max) {
+                        this.unsetFilterValue('min', true);
+                        this.unsetFilterValue('max', true);
+                    } else {
+                        this.barMinValue = e.minValue;
+                        this.barMaxValue = e.maxValue;
+                    }      
+                }, 500)
+            },
```
egcastro, line 90: Nice improvement here! I just have a few comments:
- `throttleInterval` does not seem a good name for this variable. I would call it something like `throttleId` or similar.
- I think than 500ms is way too much, we should be good with 250ms.
- I would like to avoid one liners `if`. This is a 100% preference but I prefer to always write the code the same way no matter the amount of lines of its body:
```js
if (this.throttleInterval) {
    clearInterval(this.throttleInterval);
}
```

gau94rav: Good points, thanks @egcastro, updating.

visualizations/frontend/filters/v2/TRangeInput.vue
```diff
@@ -19,6 +19,7 @@
     export default {
         data: () => ({
             is_filter: true,
+            throttleId: null,
```
gau94rav, line 22: Not sure why I missed it in first commit 😛

visualizations/frontend/filters/v2/TRangeInput.vue
```diff
@@ -76,14 +77,20 @@
...
         methods: {
             update(e) {
-                if (e.minValue === e.min && e.maxValue == e.max) {
-                    this.unsetFilterValue('min', true);
-                    this.unsetFilterValue('max', true);
-                } else {
-                    this.barMinValue = e.minValue;
-                    this.barMaxValue = e.maxValue;
+                if (this.throttleId) {
+                    clearInterval(this.throttleId);
```
sareg0, line 81: I read this about using `clearInterval` with a timeoutId

"It's worth noting that the pool of IDs used by [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) and [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) are shared, which means you can technically use clearInterval() and [clearTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout) interchangeably. However, for clarity, you should avoid doing so." - https://developer.mozilla.org/en-US/docs/Web/API/clearInterval

Maybe if you could use `clearTimeout` instead. However, I don't think you need it, unless you need to cancel the timeout for some reason.

egcastro: You want to cancel the timeout if some event is fire (basically this is the throttling behaviour we are looking for).
If you call `update`, and then you call it again before 250ms has happened, you will cancel the first setTimeout (because you do not want to execute its callback) and you will set another setTimeout...
You will keep doing this until no event is calling update within this window of 250ms.

sareg0: ah! Of course, makes sense.

gau94rav: Oh thank you @sareg0 for finding my `clearInterval` mistake 😛

visualizations/frontend/filters/v2/TRangeInput.vue
```diff
@@ -76,14 +77,20 @@
...
-                    this.barMaxValue = e.maxValue;
+                if (this.throttleId) {
+                    clearTimeout(this.throttleId);
                 }
-            }
+
+                this.throttleId = setTimeout(() => {
+                    if (e.minValue === e.min && e.maxValue == e.max) {
+                        this.unsetFilterValue('min', true);
+                        this.unsetFilterValue('max', true);
```
sareg0, line 87:
```suggestion
                        this.unsetFilterValue('min');
                        this.unsetFilterValue('max');
```

Correct me if I'm wrong, but I think that there is no longer a second argument on `unsetFilterValue`: https://github.com/snyk/topcoat-core/blob/a0eb0bf6a76bed8c7df0efe026700d4c10236b40/frontend/src/mixins/visualization_mixin.js#L301

egcastro: You're right. That's part of the old code.

sareg0 (commented):
I have a couple of thoughts that you can take or leave, depending on the problems you want to solve in this PR.  Feel free to move forward if you don't want to address any of it. 

1. I feel like a debounce function should be self-contained, and that the timer id should not be assigned to a reactive data property. Although it seems to work, it seems like it could lead to unexpected behaviour to assign the timer id to something that's reactive. 
2. ~~The input event is being fired when clicking on the bar, even though the handle does not move along the bar. This may need to be handled separately to the debounce~~ This is because the step is too small to see, but it is moving

3. I think that we should move the debounce into the computted setter, or a watcher and handling debouncing there. We should allow the barMinValue and barMaxValue to update as they please. The setting of the global state should be a side-effect, and be debounced. This is my feeling, however, feel free to correct me if this is wrong.
4. Can we use lodash? I'm not opposed to doing it yourself, but the lodash function is available to us and is re-usable. 


There is a good example of a debounce function being applied in this piece of Vue documentation: https://v2.vuejs.org/v2/guide/computed.html#Watchers

gau94rav:
> I have a couple of thoughts that you can take or leave, depending on the problems you want to solve in this PR. Feel free to move forward if you don't want to address any of it.
> 
> 1. I feel like a debounce function should be self-contained, and that the timer id should not be assigned to a reactive data property. Although it seems to work, it seems like it could lead to unexpected behaviour to assign the timer id to something that's reactive.
> 2. ~The input event is being fired when clicking on the bar, even though the handle does not move along the bar. This may need to be handled separately to the debounce~ Gaurav corrected me, this is because the step is too small to see, but it is moving
> 3. I think that we should move the debounce into the computted setter, or a watcher and handling debouncing there. We should allow the barMinValue and barMaxValue to update as they please. The setting of the global state should be a side-effect, and be debounced. This is my feeling, however, feel free to correct me if this is wrong.
> 4. Can we use lodash? I'm not opposed to doing it yourself, but the lodash function is available to us and is re-usable.
> 
> There is a good example of a debounce function being applied in this piece of Vue documentation: https://v2.vuejs.org/v2/guide/computed.html#Watchers



> I have a couple of thoughts that you can take or leave, depending on the problems you want to solve in this PR. Feel free to move forward if you don't want to address any of it.
> 
> 1. I feel like a debounce function should be self-contained, and that the timer id should not be assigned to a reactive data property. Although it seems to work, it seems like it could lead to unexpected behaviour to assign the timer id to something that's reactive.
> 2. ~The input event is being fired when clicking on the bar, even though the handle does not move along the bar. This may need to be handled separately to the debounce~ Gaurav corrected me, this is because the step is too small to see, but it is moving
> 3. I think that we should move the debounce into the computted setter, or a watcher and handling debouncing there. We should allow the barMinValue and barMaxValue to update as they please. The setting of the global state should be a side-effect, and be debounced. This is my feeling, however, feel free to correct me if this is wrong.
> 4. Can we use lodash? I'm not opposed to doing it yourself, but the lodash function is available to us and is re-usable.
> 
> There is a good example of a debounce function being applied in this piece of Vue documentation: https://v2.vuejs.org/v2/guide/computed.html#Watchers

1. A self contained timer might not work IMO, it will overwrite the state of throttle id each time update is triggered, `let timer = setTimout` will erase it, also because we have to check last state of timer, the assignment can only happen after the check before we clearTimeout.
2. Yes step is prop controller, can be handleded in page file directly.
3. You can still change the values the global state will only change once you stop dragging the pointers.
4. I wish we could use it, but we will have to global import it, which is probably not a good idea, sadly our components do not support import yet.

sareg0 (approved):

sareg0:
FWIW I think that we are globally importing lodash: https://github.com/snyk/topcoat-core/blob/f7d8174ed199322c74566d0167903bea0da79cbd/frontend/src/core.js#L154

gau94rav:
> FWIW I think that we are globally importing lodash: https://github.com/snyk/topcoat-core/blob/f7d8174ed199322c74566d0167903bea0da79cbd/frontend/src/core.js#L154
> 
> But feel free to move forward.

Oh, 😮.

No you are right, if that is already imported, I can use that.

PR #131, opened 2022-09-07
fix: Remove bottom padding from Report component

egcastro:
Get rid of that bottom padding that is producing a gap at the bottom of the table:
Screenshot 2022-09-07 at 12 46 44


gau94rav (approved):
LGTM thanks @egcastro.

PR #132, opened 2022-09-08
Cleanup files and add some defaults

sareg0:
* Add .DS_Store to .gitignore
* Remove .DS_Store file from root
* Add a .vscode folder to help with default settings across VSCode users 
  * settings.json (inspired by the settings for AppUI: https://github.com/snyk/app-ui/blob/main/.vscode/settings.json) 
  * extensions.json (inspired by the extensions recommended by AppUI: https://github.com/snyk/app-ui/blob/main/.vscode/settings.json) 

* Add eslint configuration from topcoat-core: https://github.com/snyk/topcoat-core/blob/main/frontend/.eslintrc.js

### Suggested workflow
Before starting working on a component, open it in VSCode and save the file so it is formatted. Commit it back to topcoat-public. After it is merged start work. When your work starts, you are starting with a cleanly formatted file. If you work in topcoat-builder and copy and paste code to VSCode you can save again, and it should all get formatted nicely.

Thoughts welcome.

### To Test
* point `config.yml` to `chore/cleanup-files` branch for modules in the `topcoat-builder-ide` when running locally 
```
modules:
      revision: chore/cleanup-files
```

* rebuild the modules
* Check to make sure nothing pops up in the 'topcoat-modules' folder in the the builder ide. 
* check to make sure that nothing averse happens when interaticting with pages that import topcoat-public components, and ensure that the pages render correctly.


gau94rav (approved):
Works Great!!

.eslintrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2021: true,
```
egcastro, line 4: Based on the comment on this PR: https://github.com/snyk/topcoat-core/pull/511
`eslint` should be warning us about the usage of `module.` in this file right? 
Am I missing something? Maybe I misunderstood the comment on the other PR 😄

sareg0: Indeed it should be showing an error, however it does NOT! 😓 

Weird. Do you know why there is a difference?.... Ah. I think I know. There's no `package.json`. It is failing when trying to run eslint in VSCode:

```
To use ESLint please install eslint by running npm install eslint in the workspace folder topcoat-public
or globally using 'npm install -g eslint'. You need to reopen the workspace after installing eslint.
```

I guess we need a package.json or to install all of the packages globally 🙄 
```
eslint
eslint-plugin-vue
```

I guess I could try adding a package.json and see if it is also not picked up by the builder... but it's getting mucky!

sareg0: Okay, so I've added a package.json, but it does get added to the modules folder when it is re-built

Screenshot 2022-10-19 at 16 26 57

I'm not sure what sort of consequences this has to be honest.

sareg0: It doesn't seem to add any of the dotfiles, which is interesting

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "topcoat-public",
+  "version": "1.0.0",
+  "description": "",
+  "main": ".eslintrc.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
msorber, line 7: Can we replace this script with:

`"prettify": "npx prettier --write ."`

?

gau94rav: Agree, this will be very convenient if works.

gau94rav: This will also replace the need for my PR.

sareg0: @gau94rav and @msorber I added those scripts. I tried to break them down the way they are in app-ui/product-cl though: 
https://github.com/PatchUI/ProductCL/blob/main/package.json
https://github.com/snyk/app-ui/blob/main/package.json

I also used `npm` instead of `npx` 'because'. I don't know if using `npx` is better 🤷‍♂️  I guess if we used npx we wouldn't need to install the packages. It's a tradeoff.

package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "topcoat-public",
+  "version": "1.0.0",
+  "description": "",
+  "main": ".eslintrc.js",
+  "scripts": {
+    "lint:js": "npx eslint  --color --ext .js,.json,.ts,.vue .",
+    "lint:js:fix": "npm run lint:js -- --fix",
+    "lint:format": "npx prettier . -c",
+    "lint:format:fix": "npx prettier . --write",
+    "lint:fix": "npm run lint:js:fix && npm run lint:format:fix"
```
sareg0, line 11: we could also add stylelint in the future, but I didn't want to add another thing in this PR.

gau94rav (approved):
This seems to be working great, I tested with `npm run lint:format:fix` command.

PR #134, opened 2022-09-09
Refactor pdf tooltip code.

gau94rav:
This PR has changes to allow new tooltip function.

Tooltip [PR](https://github.com/topcoat-data/topcoat-public/pull/133)

### DIFF
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -1,7 +1,7 @@
 <template>
-    <div class="relative">
+    <div class="relative" @mouseover="showTooltip = true" @mouseleave="showTooltip = false">
         <!-- PDF tooltip -->
-        <t-tooltip position="left" width="400px" is-clickable>
+        <t-tooltip position="left" width="400px" :open="showTooltip">
             <button
                 class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
                 style="height: fit-content;"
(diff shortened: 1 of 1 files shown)


library/components/TPdf.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-    <div class="relative">
+    <div class="relative" @mouseover="showTooltip = true" @mouseleave="showTooltip = false">
```
gau94rav, line 2: Events have strange behaviour when used on button, using in container div works fine.

sareg0: interesting, like what?

gau94rav: For me, it was not even triggering these events for some reason, even though it should tbh, not sure what caused it, not a huge concern in this case, but this is something I would like to check deeply, what did I miss.

sareg0: I tested putting the mouseover/leave on the button and it  seemed to work 🤷 

I noticed though that the tooltip is not very easy to keep open because it is placed a bit away from the button, it is hard to hover over

sareg0 (approved):

library/components/TPdf.vue
gau94rav: In this case maybe it's not going to matter to keep it open, but I see what you mean, in some cases there might be a link to click inside, this is something to fix inside TTooltip component, I'll take a look, thanks.

PR #135, opened 2022-09-09
Refactor/t-tooltip

gau94rav:
Remove un-necessary code and clean methods.

### DIFF
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -2,7 +2,6 @@
 	<div
         @mouseover="handleMouseOver"
         @mouseleave="handleMouseLeave"
-        @click="handleClick"
         ref="tooltipContainer"
     >
 		<div ref="trigger">
@@ -16,7 +15,7 @@
 			class="p-4 bg-[#1C1C21] text-white w-max h-max rounded fixed z-50"
 			:style="{ width, ...positions.tooltip }"
 			ref="tooltip"
-			v-show="showTooltip && $slots.default"
+			v-show="isVisible && $slots.default"
 		>
 			<slot></slot>
 			<div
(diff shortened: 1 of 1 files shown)


library/components/TTooltip.vue
```diff
@@ -38,17 +37,15 @@ export default {
 			type: String,
 			default: "top"
 		},
+		open: { // To control state from a component.
```
egcastro, line 40: I'd like to use `isOpen` for boolean states. Summarised how we can named boolean props:

Use prefix is, can, has:

-  is: for visual/behavior variations. e.g. isVisible, isEnable, isActive
- can: fore behavior variations or conditional visual variations. e.g. canToggle, canExpand, canHaveCancelButton
- has: for toggling UI elements. e.g. hasCancelButton, hasHeader

library/components/TTooltip.vue
```diff
@@ -83,27 +80,36 @@ export default {
...
+			this.isVisible = true;
 		},
 		hide() {
-			this.showTooltip = false;
-            this.isFocused = false;
+			this.isVisible = false;
+		}
+	},
+	watch: {
+		open(isOpen) {
```
egcastro, line 108: I couldn't see from where this `isOpen` comes from

gau94rav: You the parameter? the prop `open` is being watched and it's value `isOpen` can be true or false.

library/components/TTooltip.vue
gau94rav: Yes, probably best to use that prefix because it is just like our other components, @sareg0 agree?

sareg0: sounds good

library/components/TTooltip.vue
```diff
@@ -83,27 +80,36 @@ export default {
...
+			this.isVisible = false;
+		}
+	},
+	watch: {
+		isOpen(open) {
+			if (open) {
+				return this.show();
+			}
+			return this.hide();
 		}
```
sareg0, line 113: I don't think you need this watcher

library/components/TTooltip.vue
```diff
@@ -38,17 +37,15 @@ export default {
...
+		isOpen: { // To control state from a component.
+			type: Boolean,
+			default: false,
+		}
 	},
 	data: () => ({
 		positions: {},
-		showTooltip: false,
-        isFocused: false,
+		isVisible: false,
```
sareg0, line 47: Can you do this instead and skip the watcher?
```suggestion
		isVisible: isOpen,
```

gau94rav: If we use this method, then placement logic might not work and depending on window size and position of elements at that point, tooltip will not be able to align itself properly.

I am sure placement logic can be handled better, though it requires a lot more time I think to change.

gau94rav: Watcher makes sure that placement is also triggered on change.

gau94rav: Oh wait, sorry I forgot that I changed that, you are right, trying this approach.

sareg0 (approved):

PR #136, opened 2022-09-12
Handle parent context for copying url.

gau94rav:
This PR will add support for `context[parent]` inside url copy component.

**How to test**
1. Switch to this branch inside IDE.
2. Build modules.
3. Go to executive page and click Copy URL button.

**Expected results**
1. If opened inside snyk website, the url matches with app.snyk.io
2. If opened outisde snyk website, the url matches with current url.
3. Also copies all filters.

### DIFF
--- a/library/components/TUrlCopyButton.vue
+++ b/library/components/TUrlCopyButton.vue
@@ -1,43 +1,72 @@
 <template>
-    <div class="relative">
-        
-        <t-tooltip position="bottom" width="130px">
-            <button
-                class="h-[30px] w-[30px] rounded-[4px] flex items-center justify-center border border-[#C3C2CB] p-1 transition-colors"
-                @click="copy"
-                :class="{ 'bg-[#B4E4D9]': copied }"
-                slot="trigger"
-            >
-                <check-icon :size="18" v-if="copied" />
-                <link-icon :size="18" v-else />
-                <span class="visually-hidden">{{ copied ? "URL Copied" : "Copy URL" }}</span>
-            </button>
-            {{ copied ? "URL Copied" : "Copy URL" }}
-        </t-tooltip>
-    </div>
+  <div class="relative">
+    <t-tooltip position="bottom" width="130px">
+      <button
+        class="h-[30px] w-[30px] rounded-[4px] flex items-center justify-center border border-[#C3C2CB] p-1 transition-colors"
+        @click="copy"
+        :class="{ 'bg-[#B4E4D9]': copied }"
+        slot="trigger"
+      >
+        <t-loading-spinner position="relative" v-if="!whoami" />
+        <check-icon :size="18" v-else-if="copied" />
+        <link-icon :size="18" v-else />
+        <span class="visually-hidden">{{
+          copied ? "URL Copied" : "Copy URL"
+        }}</span>
+      </button>
+      {{ copied ? "URL Copied" : "Copy URL" }}
+    </t-tooltip>
+  </div>
 </template>
 
 <script>
-    export default {
-        data: () => ({
-            copied: false,
-        }),
-        methods: {
-            copy() {
-                this.copied = true;
-                const tempInput = document.createElement('input'),
-                text = window.location.href;
-                document.body.appendChild(tempInput);
-                tempInput.value = text;
-                tempInput.select();
-                document.execCommand('copy');
-                document.body.removeChild(tempInput);
-                setTimeout(() => {
-                    this.copied = false;
-                }, 1000)
-            }
+export default {
+  data: () => ({
+    copied: false,
+  }),
+  computed: {
+    url() {
+      if (this.parentUrl) {
+        this.parentUrl.searchParams.set("context[page]", this.page.url);
+        for (const [filter, value] of Object.entries(this.filters)) {
+          this.parentUrl.searchParams.set(filter, encodeURIComponent(value));
         }
-    }
+        return this.parentUrl.toString();
+      }
+
+      return window.location.href;
+    },
+    parentUrl() {
+      try {
+        if (this?.whoami?.context?.parent) {
+          return new URL(this.whoami.context.parent[0]);
+        }
+      } catch (e) {
+        console.log(e);
+      }
+      return null;
+    },
+  },
+  methods: {
+    copy() {
+      if (!this.whoami) return;
+      this.copied = true;
+
+      const tempInput = document.createElement("input");
+      document.body.appendChild(tempInput);
+      tempInput.value = this.url;
+      tempInput.select();
+
+      document.execCommand("copy");
+      document.body.removeChild(tempInput);
+
+      setTimeout(() => {
+        this.copied = false;
+        // Timeout to show success message and hide it.
+      }, 1000);
+    },
+  },
+};
 </script>
 
 <style>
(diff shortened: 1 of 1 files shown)


library/components/TUrlCopyButton.vue
```diff
@@ -1,57 +1,75 @@
...
+        if (parentUrl) {
+          const pageContext = 'context[page]=' + this.page.url;
+          return parentUrl + "?" + pageContext + "&" + this.urlParams;
         }
+
+        return window.location.href;
+      },
+      urlParams() {
+        return new URLSearchParams(this.filters).toString();
+      }
```
gau94rav, line 38: Due to formatting, it might be hard to see changes, I am highlighting the parts that are changed.

library/components/TUrlCopyButton.vue
```diff
@@ -1,57 +1,75 @@
...
+        tempInput.value = this.url;
+        tempInput.select();
+
+        document.execCommand('copy');
+        document.body.removeChild(tempInput);
+
+        setTimeout(() => {
+          this.copied = false;
+          // Timeout to show success message and hide it.
+        }, 1000)
```
gau94rav, line 55: Also some changes here to target computed property as url.

library/components/TUrlCopyButton.vue
```diff
@@ -1,57 +1,75 @@
...
+    data: () => ({
+      copied: false,
+    }),
+    computed: {
+      url() {
+        const parentUrl = this.filters['context[parent]'];
+
+        if (parentUrl) {
+          const pageContext = 'context[page]=' + this.page.url;
+          return parentUrl + "?" + pageContext + "&" + this.urlParams;
```
sareg0, line 31: Is there a possibility to push this calculation into the mixins (either now or as a followup). It would help keep information about `context` params. Out of topcoat-public component, and centralise it for re-use by other components that rely on the parent url

library/components/TUrlCopyButton.vue
```diff
@@ -1,57 +1,75 @@
...
+    methods: {
+      copy() {
+        this.copied = true;
+
+        const tempInput = document.createElement('input');
+        document.body.appendChild(tempInput);
+        tempInput.value = this.url;
+        tempInput.select();
+
+        document.execCommand('copy');
```
egcastro, line 49: This is kind of old API to create this kind of interactions. We should be using the new clipboard API instead: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard

Here we have an example: https://googlechrome.github.io/samples/async-clipboard/

And the API is Promise based so we can set the `copied` attribute in the `then` instead of using the `setTimeout` that you added on line 52

library/components/TUrlCopyButton.vue
egcastro, line 31: We should not append query string parameters and concatenated them by hand. We can use a combination of `URL` and `URLSearchParams` API instead and use `set` method to add parameter or create a full URL.

library/components/TUrlCopyButton.vue
gau94rav: @egcastro Clipboard API is blocked under iframe, due to security permission policy.

sareg0 (approved):
The formatting is looking soooo much better!! 💅 💅 💅 

This works in the 'old Chrome'. I tested in the new chrome to confirm that it's not working there. We can re-test it when Emi merges https://github.com/snyk/app-ui/pull/480

library/components/TUrlCopyButton.vue
gau94rav: Sounds good, though need to think about adding it to library_mixin, might be out of place.

gau94rav: You mean, loop the filters object and `set` url params that way?

gau94rav:
> 

Maybe i am missing something, how can I test in new chrome as well?

library/components/TUrlCopyButton.vue
sareg0: There are some good examples of that here? https://github.com/snyk/topcoat-core/blob/b0e61e137925e8427b3df162451d94823c688a25/frontend/src/store/modules/page.js

sareg0:
> > 
> 
> @sareg0 Maybe i am missing something, how can I test in new chrome as well?

You'd need the feature flags on, and Emi's PR will need to be merged first. We can set you up with the correct feature flags when his PR is merged and we can test again.

library/components/TUrlCopyButton.vue
egcastro: Ok you're right. So the way we should be doing this is by creating a new message and `postMessage` a command that says `copy` with the string we want to be copied and we should use the clipboard API in the parent iframe.

If we are not willing to do this because of any reason, we should check browser compatibility because the execCommand API is deprecated and it has poor cross-browser support.

image

gau94rav: I see, let's go for this approach for now, then we can create the PR in registry to handle such operation.
Then it should be like this.
1. Copy base url if not in app.snyk.io
2. Send message to parent if there is a context.

This might reduce the url size because then we do not need a full url as context, instead it can be just `context[parent]=true`

egcastro: Well, but to keep it consistent we should have a method that hides the logic for components. We can create the method `copyToClipboard` for instance in the visualization_mixin.
There we should use the new `Clipboard` API if we are not inside an iframe or use the `postMessage` otherwise.
This way we hide all the logic of `iframe` vs `not-iframe` out of components. So that it's easier and centralised when we have a version that can run outside of iframes.

gau94rav: In this particular case, the component is library component, and sure other components might use copy logic as well, so probably makes sense to take it out of this component and copy to `common_mixin`, I'll see if it works good that way and split the PR.

Because Ezra probably needs this merged today, we can delay postMessage change for now.

gau94rav: Got it, in that case let's leave everything in this component for now 😅, we will move to core with final version of this function. Thanks.

library/components/TUrlCopyButton.vue
```diff
@@ -24,17 +24,24 @@
...
-        return new URLSearchParams(this.filters).toString();
+      parentUrl() {
+        const parentUrl = this.filters['context[parent]'];
+        try {
+          return new URL(parentUrl);
+        } catch (e) {
+          console.log(e);
+        }
+        return null;
       }
```
gau94rav, line 45: @egcastro this is the correct approach?

library/components/TUrlCopyButton.vue
```diff
@@ -24,17 +24,24 @@
...
-
-        if (parentUrl) {
-          const pageContext = 'context[page]=' + this.page.url;
-          return parentUrl + "?" + pageContext + "&" + this.urlParams;
+        if (this.parentUrl) {
+          this.parentUrl.searchParams.set("context[page]", this.page.url);
+          for (const [filter, value] of Object.entries(this.filters)) {
+            this.parentUrl.searchParams.set(filter, encodeURIComponent(value));
+          }
+          return this.parentUrl.toString();
```
gau94rav, line 32: @egcastro this is the correct approach for URL?

egcastro:
This PR needs to be updated with the new context store added in this other PR: https://github.com/snyk/topcoat-core/pull/512

sareg0 (approved):

PR #137, opened 2022-09-12
Fix/tooltip transition

gau94rav:
**Problem:**
1. When tooltip opens, it sometimes have a broken layout for half a second, until it corrects itself.
2. Transition of tooltip open and close is very static.

**Solution**
1. Prevent tooltip from resetting it's position state on each mouse hover and trigger placement logic AFTER it appears.
2. Add fade transition using vue's `<transition />` tag.

**How to test**
1. Switch to this branch inside IDE.
2. Open executive or issues page, click on pdf download or hover on url copy button.

**Expected results**
1. No broken flash when it opens.
2. Subtle fade animation on open and close.

### DIFF
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -13,7 +13,7 @@
 		<!-- $slots.default is used in cases when the trigger is using v-if on content's div -->
 		<transition name="fade">
 			<div
-				class="p-4 bg-[#1C1C21] text-white w-max h-max rounded fixed z-50"
+				class="p-2 bg-[#1C1C21] text-white w-max h-max rounded fixed z-50"
 				:style="{ width, ...positions.tooltip }"
 				ref="tooltip"
 				v-show="isVisible && $slots.default"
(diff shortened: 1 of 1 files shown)


library/components/TTooltip.vue
```diff
@@ -11,18 +11,20 @@
...
 
 		<!-- $slots.default is used in cases when the trigger is using v-if on content's div -->
-		<div
-			class="p-4 bg-[#1C1C21] text-white w-max h-max rounded fixed z-50"
-			:style="{ width, ...positions.tooltip }"
-			ref="tooltip"
-			v-show="isVisible && $slots.default"
-		>
-			<slot></slot>
+		<transition name="fade">
```
gau94rav, line 14: Change for transition.

sareg0 (approved):

sareg0:
Thanks for leaving the comment about the code change!
I wouldn't be opposed to fixing the indentation of this file if you felt like doing that too, but it doesn't have to block you.

PR #141, opened 2022-09-16
correction in tooltip placement logic

gau94rav:
Tooltip has multiple behavioural issues when it comes to conditionals and placements.

**Issues Fixes:**
1. Fails to place itself according to width and height of trigger until mouse moves again.
2. Can't keep open for user to click a link inside tooltip.

Jira [ticket](https://snyksec.atlassian.net/jira/software/c/projects/DP/boards/443?modal=detail&selectedIssue=DP-377&assignee=62261dde94f7e200690091ae)

One example with PDF component.
**Before**
Screenshot 2022-09-16 at 6 56 01 PM

**After**
Screenshot 2022-09-19 at 2 17 11 PM

### DIFF
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -5,7 +5,7 @@
     @mouseleave="showTooltip = false"
   >
     <!-- PDF tooltip -->
-    <t-tooltip position="left" width="400px" :open="showTooltip">
+    <t-tooltip position="left" width="400px" :open="is_loading && showTooltip">
       <button
         class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
         style="height: fit-content"
--- a/library/components/TPdf.vue
+++ b/library/components/TPdf.vue
@@ -5,7 +5,11 @@
     @mouseleave="showTooltip = false"
   >
     <!-- PDF tooltip -->
-    <t-tooltip position="left" width="400px" :open="is_loading && showTooltip">
+    <t-tooltip
+      position="left"
+      width="400px"
+      :is-open="is_loading && showTooltip"
+    >
       <button
         class="bg-[#145DEB] border-[#145DEB] text-white text-sm px-2 py-[5px] rounded-[4px]"
         style="height: fit-content"
--- a/library/components/TTooltip.vue
+++ b/library/components/TTooltip.vue
@@ -1,10 +1,11 @@
 <template>
-  <div ref="tooltipContainer">
-    <div
-      ref="trigger"
-      @mouseleave="handleMouseLeave"
-      @mouseover="handleMouseOver"
-    >
+  <div
+    ref="tooltipContainer"
+    class="relative w-max"
+    @mouseleave="handleMouseLeave"
+    @mouseover="handleMouseOver"
+  >
+    <div ref="trigger">
       <slot name="trigger">
         <help-circle-outline-icon :size="12" />
       </slot>
@@ -12,9 +13,9 @@
 
     <!-- $slots.default is used in cases when the trigger is using v-if on content's div -->
     <transition name="fade">
-      <div class="relative w-0 h-0" :style="{ ...positions.tooltip }">
+      <div class="absolute" :style="{ ...styles.tooltip }">
         <div
-          class="p-2 bg-[#1C1C21] text-white text-sm w-max h-max rounded absolute z-50"
+          class="p-2 bg-[#1C1C21] text-white text-sm w-max h-max rounded relative z-50"
           :style="{ width }"
           ref="tooltip"
           v-show="isVisible && $slots.default"
@@ -24,7 +25,7 @@
           <slot></slot>
           <div
             class="bg-[#1C1C21] w-3 h-3 absolute rotate-45"
-            :style="{ ...positions.arrow }"
+            :style="{ ...styles.arrow }"
           ></div>
         </div>
       </div>
@@ -50,7 +51,7 @@ export default {
     },
   },
   data: () => ({
-    positions: { arrow: {}, tooltip: {} },
+    styles: { arrow: {}, tooltip: {} },
     isVisible: false,
     isTooltipFocused: false, // If cursor is on tooltip.
   }),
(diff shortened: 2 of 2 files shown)


library/components/TTooltip.vue
```diff
@@ -1,28 +1,32 @@
...
     <!-- $slots.default is used in cases when the trigger is using v-if on content's div -->
     <transition name="fade">
-      <div
-        class="p-2 bg-[#1C1C21] text-white text-sm w-max h-max rounded fixed z-50"
-        :style="{ width, ...positions.tooltip }"
-        ref="tooltip"
-        v-show="isVisible && $slots.default"
-      >
-        <slot></slot>
+      <div class="relative w-0 h-0" :style="{ ...positions.tooltip }">
```
gau94rav, line 15: Change 1:
Having a parent div that has no height or width but relative position will make sure we do not have keep the exact position inside DOM for tooltip placement, it always stays near trigger.

library/components/TTooltip.vue
```diff
@@ -33,20 +37,22 @@ export default {
...
+      // To control state from a component.
       type: Boolean,
       default: false,
-    }
+    },
   },
   data: () => ({
     positions: { arrow: {}, tooltip: {} },
     isVisible: this.isOpen,
+    isTooltipFocused: false, // If cursor is on tooltip.
```
gau94rav, line 55: Change 2:
This will allow us to keep our tooltip opened if user moves cursor on it.

library/components/TTooltip.vue
```diff
@@ -56,27 +62,31 @@ export default {
...
+          positions.arrow.top = "6px";
         } else {
-          positions.tooltip.top = position.top - tooltipElement.clientHeight / 2 + "px";
-          positions.tooltip.left = position.left + triggerElement.clientWidth + 15 + "px";
-          positions.arrow.top = position.top + "px";
-          positions.arrow.left = position.left + triggerElement.clientWidth + 10 + "px";
+          positions.tooltip.top = "-20px";
+          positions.tooltip.left = triggerElement.clientWidth + 7 + "px";
+          positions.arrow.left = "-6px";
+          positions.arrow.top = "5px";
```
gau94rav, line 89: Change 3:
More solid placement values around trigger, thanks to relative position of parent, this will always be accurate placement.

library/components/TTooltip.vue
```diff
@@ -93,29 +103,41 @@ export default {
       }
       this.hide();
     },
+    handleTooltipFocus(isFocused) {
+      this.isTooltipFocused = isFocused;
+      if (!isFocused) {
+        this.hide();
+      }
+    },
```
gau94rav, line 111: Change 4:
When user moves away the cursor from tooltip.

library/components/TTooltip.vue
```diff
@@ -93,29 +103,41 @@ export default {
...
+    hide: window._.debounce(function () {
+      // Debounce will give user chance to hover to tooltip and keep it open, as long as cursor stays on it.
+      // This is useful when tooltip contains clickable links.
+      if (this.isTooltipFocused) {
+        // If user is interacting with tooltip
+        return;
+      }
       this.isVisible = false;
-    }
+    }, 100),
```
gau94rav, line 128: Change 5: 
Gives enough chance to user to move to tooltip and keep it open.

library/components/TTooltip.vue
```diff
@@ -93,29 +102,51 @@ export default {
...
+  },
+  watch: {
+    // Watcher triggers placement logic as well.
+    isOpen(n) {
+      if (n) {
+        this.show();
+      } else {
+        this.hide();
+      }
+    },
```
gau94rav, line 137: Change 6:
Without watcher there is no way to re-trigger the placement when open state is controlled by parent trigger.

library/components/TPdf.vue
```diff
@@ -5,7 +5,11 @@
     @mouseleave="showTooltip = false"
   >
     <!-- PDF tooltip -->
-    <t-tooltip position="left" width="400px" :open="showTooltip">
+    <t-tooltip
+      position="left"
+      width="400px"
+      :is-open="is_loading && showTooltip"
+    >
```
gau94rav, line 12: Missed renaming prop in last PR.

egcastro:
May I request some `before and after` screenshots of the fixes you've applied? It nice to have that as a reference in the PR description to see visually what has changed before testing the code itself.

gau94rav:
> May I request some `before and after` screenshots of the fixes you've applied? It nice to have that as a reference in the PR description to see visually what has changed before testing the code itself.

Sure, I have added one example for it, thanks.

egcastro:
I think we should improved the positioning a little bit more. It should be centered. Somewhere similar to the red line here:  
tooltip

gau94rav:
> I think we should improved the positioning a little bit more. It should be centered. Somewhere similar to the red line here: tooltip
Corrected placement of pdf tooltip.
Screenshot 2022-09-19 at 2 17 11 PM

sareg0 (commented):
Perhaps we could think about allowing the tooltip some extra area, so that the mouseover/mouseleave works well

if you take a look at the tooltips in Product CL, they keep a surrounding area, so that if you hover towards the tooltip the tooltip mouseover/leave events take over, even if you are not hoering the tooltip area with the message on it

Screenshot 2022-09-20 at 11 26 52
The blue area indicates the tooltips hoverable area. It makes the tooltip hover very forgiving. 
https://productcl.netlify.app/?path=/docs/components-info-badge--tooltip

library/components/TTooltip.vue
sareg0: Non-blocking: I preferred it when we had one data property that reflects the current visibility of the component. It felt like that was a single source of truth. Now it feels like there are two, which makes it harder to reason about the component.

library/components/TTooltip.vue
```diff
@@ -33,53 +38,58 @@ export default {
...
+          styles.arrow.right = "-6px";
+          styles.arrow.top = "10px";
         } else {
-          positions.tooltip.top = position.top - tooltipElement.clientHeight / 2 + "px";
-          positions.tooltip.left = position.left + triggerElement.clientWidth + 15 + "px";
-          positions.arrow.top = position.top + "px";
-          positions.arrow.left = position.left + triggerElement.clientWidth + 10 + "px";
+          styles.tooltip.top = "0px";
+          styles.tooltip.left = triggerElement.clientWidth + "px";
+          styles.tooltip.paddingLeft = "10px";
```
gau94rav, line 87: With a bit of tweaking and using paddings, I was able to get rid of debounce and make tooltips hoverable.

sareg0: amazing!

PR #142, opened 2022-09-16
Server sort

msorber:
Goal: provide server-side sorting that is done in the sql to the table.

To test:
1. In  snyk-insights, check out the "server_sort" branch and run "build and sync modules"
2. Open snyk-insights in the [iFrame](https://app.snyk.io/group/5ea22317-e879-4eb9-8ef7-411dc7c3a604/reporting?context%5Bpage%5D=issues-detail&issue_status=Open&issue_by=Severity)
3. Verify that the table starts sorted by "SCORE" descending, "ISSUE" ascending, and that the "PRODUCT" column is sortable but not currently sorted.
4. Click the up/down/up-down icons to sort by columns
5. Open the URL in a new tab and verify that the updated sorting is preserved- not just which columns are sorted, but also the order in which they are sorted
6. Export a pdf and verify that the sorted table in the pdf matches the page
7. Export csv and verify that the rows are sorted in the same way as the table.

Known issues: 
The table has grown into a cluster, I'm aware that the sorting data is stored in 4 places which is _ridiculous_. But the correct fix for that is a complete refactor of the table to split the layer handling into a wrapper and turn the table back into a "dumb" non-topcoat component, plus a refactor of the export to csv component- both of which are far outside the scope of this PR. I will tackle those refactors ASAP, so if you see anything here please feel free to call it out- I will make a note of it even if it falls outside of the scope of this sorting PR.

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -1,22 +1,21 @@
 <template>
   <div class="rootTableContainer">
     <slot name="columnConfig" :setColumnConfig="setColumnConfig"></slot>
-
       <!-- Title -->
       <div class="tableHeaderContainer">
         <div class="title">
-            <span v-if="title">{{ title }}</span>
+            <span v-if="title">{{ title }} </span>
         </div>
 
         <div class="tableControls" v-show="!isPdf">
           <!-- <div>Future Group By</div> -->
-          <t-select-columns
+          <test-select-columns
             v-if="modifiableColumns"
             :modifiableColumns="modifiableColumns"
             @updateFilteredColumns="updateFilteredColumns"
             :t-layer="layer"
           />
-          <TCsvExport
+          <t-csv-export
             v-if="enableCsvDownload"
             :t-layer="layer"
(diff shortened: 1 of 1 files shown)


sareg0 (approved):

visualizations/frontend/other/TTable.vue
```diff
@@ -61,25 +61,23 @@
...
-          <slot
-            v-if="arrowDirection[column.property] === 'asc'"
-            name="sortAscendingIcon"
-            v-bind="column"
-          >
-            <div class="icon move_up">&#8964;</div>
-          </slot>
-          <slot v-else name="sortDescendingIcon" v-bind="column">
-            <div class="icon move_down">&#8963;</div>
+        <div style="display: flex;align-items: center;">
```
sareg0, line 64: Is there a reason you aren't using tailwind classes here?

sareg0 (approved):
I noticed a few things when testing. Where would you like them to be captured?

msorber:
> I noticed a few things when testing. Where would you like them to be captured?

Yes please!

PR #143, opened 2022-09-19
remove underline from links

msorber:
Goal: remove the underlines from links, per conversation with @gau94rav 

To test: in snyk-insights check out the "remove_link_underline" branch and sync modules, then open the issue details page and verify that all the links are no longer underlined.


gau94rav (approved):
LGTM, thanks @msorber

PR #144, opened 2022-09-20
fix: Add virtualization to TMultiSelector

egcastro:
Use the new component added in: https://github.com/snyk/topcoat-core/pull/516 to support big data sets.
This changes is being applied only to the `TMultiSelector` component but it should be easy to adapt it to other if needed.
Use the new added method `getColumnsByNameAndAttribute` in the PR mentioned above to avoid looping twice over the whole `rows` data set.
Fixed some UI layout issues.

Related JIRA ticket: https://snyksec.atlassian.net/browse/DP-271?atlOrigin=eyJpIjoiOGVjMWFjNDAzNzhhNGYzM2E0NGJmZjc0MDYwNDFmOGEiLCJwIjoiaiJ9

### DIFF
--- a/visualizations/frontend/filters/v2/TMultiSelector.vue
+++ b/visualizations/frontend/filters/v2/TMultiSelector.vue
@@ -22,13 +22,24 @@
 
 		<!-- Popup Contents -->
 		<div class="min-w-[294px]" v-bind="popupAttrs">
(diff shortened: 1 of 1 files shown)


gau94rav (approved):
LGTM

PR #149, opened 2022-09-27
show relative preset on date button if selected

gau94rav:
When a preset is selected instead of custom absolute date, show that selected preset as filter label.

[Design](https://www.figma.com/file/IaywW74sLSgYJSYHjOC0ZL/Reporting-Beta-2022)

### DIFF
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -12,7 +12,19 @@
       @confirm="handleConfirm"
       @pick="setCustomPreset"
       @open="fetchLayerData"
+      :class="relativePreset && 'w-auto'"
     >
+      <div
+        slot="input"
+        v-if="relativePreset"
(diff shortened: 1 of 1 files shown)


sareg0:
@gau94rav can you can a screenshot of the design I should look at? I am getting taken to the wider file and I'm having trouble finding the calendar component.

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -27,18 +39,21 @@
         </button>
       </div>
 
+      <!-- To hide default clear icon -->
```
sareg0, line 42: Is that comment accurate? I

gau94rav: Yes I am using `<div slot="icon-clear"></div>` to hide the default icon of this component, it does not match with our design.

gau94rav:
> Design

@sareg0 The actual calendar component is not really present in our figma design but we have this component that should show up when a preset is selected instead of an absolute date.

Screenshot 2022-09-28 at 2 11 00 PM

PR #150, opened 2022-09-27
chore: use chevron instead of +/- for row toggle controls

sareg0:


visualizations/frontend/other/TTable.vue
```diff
@@ -125,18 +125,15 @@
...
-                  <div class="icon move_up">+</div>
-                </slot>
-              </div>
-            </div>
+            <button v-if="canCollapseDetailRows" @click="toggleRow(row)">
+              <span aria-hidden="true">
+                <chevron-down-icon v-if="row.detailRowOpen"  />
+                <chevron-right-icon v-else />
+              </span>
+              <span class="sr-only">
```
sareg0, line 133: This is a tailwind class. We may be able to replace the visually-hidden class in a different component: https://tailwindcss.com/docs/screen-readers

msorber (approved):
LGTM!

visualizations/frontend/other/TTable.vue
msorber: That is cool! Is there a way to link the screen reader text to the icons themselves rather than having a different span? Wouldn't a SR have to tab through the icon to get to the text?

sareg0: Indeed. My first implementation looked like the following.

```js
<button>
  <chevron-down-icon v-if="row.detailRowOpen" aria-label="collapse row" /> 
  <chevron-right-icon v-else aria-label="open row" />
</button>
```
However, the `aria-label` attributes didn't seem to fall through, so it still yielded `aria-label="Chevron down icon"`. Maybe it needs to be a prop?  There's not much information in the Icons repo https://github.com/PatchUI/Icons

The above would be sufficient if the aria-label worked. We could wrap the icons in an element that has an aria-label on it, but it seemed as efficient to use add the text to the button, which is more of a button thing to do anyway.

>  Wouldn't a SR have to tab through the icon to get to the text?
It's a good question. 

In this case, once wrapped with `aria-hidden` it's as if the icons don't exist to the screen reader. So when they tab to the button only the text is read out, and any movement through the DOM won't pick up the icon elements.

msorber (approved):

PR #156, opened 2022-09-29
Table updates for Grouped issues table

sareg0:
This is all @icoconutgirl 

This PR makes some fixes for the table design as per the following Jira ticket

Jira ticket: https://snyksec.atlassian.net/browse/DP-399

### If you want to test

The grouped table is being used in the Vulnerability Report

* Switch the Git branch in https://snyk-insights.topcoatdata.app/a/develop to point to `feat/vulnerability`
* change the https://snyk-insights.topcoatdata.app/a/develop/project/config|yml] to point to `chore/table-updates`

```
modules:
      revision: chore/table-updates
```
* rebuild modules, by clicking the cloud icon.
* Go to https://snyk-insights.topcoatdata.app/vuln-detail You should see the grouped table there.

### Before Changes
Screen Shot 2022-09-29 at 13 51 10-fullpage

### After Changes
Screen Shot 2022-09-29 at 13 50 28-fullpage

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -3,7 +3,7 @@
     <slot name="columnConfig" :setColumnConfig="setColumnConfig"></slot>
 
       <!-- Title -->
-      <div class="tableHeaderContainer">
+      <div class="tableHeaderContainer" v-if="showTableHeader">
         <div class="title">
             <div v-if="title">{{ title }}</div>
             <t-tooltip v-if="tooltip" position="top" class="tooltip" width="260px">
@@ -180,6 +180,7 @@
               v-for="(column, cindex) in internalColumns"
               :key="column.property"
               :ref="'rowCell_' + gindex + '_' + rindex + '_' + cindex"
+              class="border-b border-[#D3D3D9] align-top py-[12px]"
               :class="generateCellClasses({ column, cindex, rindex })"
             >
               <slot
@@ -199,7 +200,9 @@
               :key="'detailRow' + rindex"
               class="spanAllColumns"
             >
-              <slot name="detail_row_slot" v-bind="row.originalRow"></slot>
+              <div class="px-4">
+                <slot name="detail_row_slot" v-bind="row.originalRow"></slot>
+              </div>
             </div>
           </div>
         </div>
@@ -313,7 +316,7 @@ export default {
     cellClasses: {
       type: String,
       default:
-        "border-b border-[#D3D3D9] align-top pt-[12px] pb-[17px] snykCell table-data",
+        "break-all",
     },
     layerColumnsToHide: {
       type: Array,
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -3,7 +3,7 @@
     <slot name="columnConfig" :setColumnConfig="setColumnConfig"></slot>
 
       <!-- Title -->
-      <div class="tableHeaderContainer">
+      <div class="tableHeaderContainer" v-if="title || tooltip  || enableCsvDownload || canSearch || (modifiableColumns && modifiableColumns.length > 0)">
```
sareg0, line 6: When I saw this I thought 'oh my, what will happen when we need to add another piece to this condition?'. 
We could move this to a computed property, however, we could also take a look at how we are using spacing on elements to create layout, and adjust things so that if an element isn't rendered it takes no space: 

I created an RFC to talk about this: https://github.com/topcoat-data/topcoat-public/pull/157

visualizations/frontend/other/TTable.vue
```diff
@@ -313,7 +315,7 @@ export default {
     cellClasses: {
       type: String,
       default:
-        "border-b border-[#D3D3D9] align-top pt-[12px] pb-[17px] snykCell table-data",
+        "border-b border-[#D3D3D9] align-top py-[12px] break-all",
```
sareg0, line 318: The effect of using `break-all` is slightly different to `word-break: break-word`. In addition `break-words` (tailwind class), does not break the words as we would like. `break-all` seems to come the closest among the Tailwind utilities.  However, [MDN documentation on word-break ](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) suggests we want the following to get the exact behaviour of `word-break: break-all;` 


```
word-break: normal;
overflow-wrap: anywhere;
```

Tailwind doesn't seem to have anything that matches this exactly. So maybe we create our own utility by renaming `snykCell` to something else utility like, or we just go with break-all and accept the differences for now.

visualizations/frontend/other/TTable.vue
msorber: Your RFC is fantastic, that is a MUCH better approach.

visualizations/frontend/other/TTable.vue
```diff
@@ -374,14 +376,17 @@ export default {
...
   computed: {
-      isPdf(){
-          return window.location.href.includes('/pdf?')
-      },
-      allInitializationComplete(){
-        const isPagingInitialized = !this.canPageServer || this.endIndex
-        const isModifiableColumnsInitialized = !this.modifiableColumns || this.$store.state.layers.components[this.tag_unique]?.filters?.column_list !== undefined
-        return isPagingInitialized && isModifiableColumnsInitialized
-      },
+    showTableHeader() {
```
sareg0, line 379: maybe `hasTableHeader`? 🤷

visualizations/frontend/other/TTable.vue
msorber: The snykCell with "word-break: break-word" was added to address an issue with very long project's forcing the column with the project to take up 60+% of the width because word-break: normal had no place to break the name up, so break-all should work just fine.

However now that I look at this again, the default styling was added here and made overridable because we weren't sure that it made sense to force the snyk table styling on every table instance, but now that we have been here a while I think it does make sense to have that styling be universal. Except for the break-word/all css which should only be applied to the issues detail table.

So I propose

1. removing snykCell and adding cellClass="break-all" to the issues detail table in both the page and pdf template
2. move "border-b border-[#D3D3D9] align-top py-[12px]" from the prop to class="border-b border-[#D3D3D9] align-top py-[12px]" on ~line 183 below or above where "generateCellClasses" is called.

Thoughts? If we agree I can make those changes real quick.

visualizations/frontend/other/TTable.vue
msorber: If your updates in chore/table-gaps-refactoring make this a non-issue I vote for that approach instead.

msorber (dismissed):
See comments please!

visualizations/frontend/other/TTable.vue
sareg0: I left `cellClasses` as is, to leave the props interface to the outside intact. I made the other changes. Though

msorber (approved):

visualizations/frontend/other/TTable.vue
msorber: 👍

msorber (approved):

PR #157, opened 2022-09-29
RFC: rejig how spacing is created between TTable sections

sareg0:
This is a proposal to rejig how we space items in the table, particularly the header. We had to come up with a big long conditional to hide a piece of the UI in this PR: https://github.com/topcoat-data/topcoat-public/pull/156/files#diff-7e4511d830c406b3a4cb3b11b35473535e081cec4d6280eef0e1c7c991f4a474R6

Given that Vue doesn't render components with `v-if`, if we add `gap` on a flex container, it will space its flex items nicely. It means we don't need to rely on the spacing on an individual item to create the spacing needed, so if it isn't rendered the spacing will be predictable.

This isn't a comprehensive refactoring of the spacing and is only intended to enable the aforementioned PR to avoid a big conditional. However, if we like this approach we could look for more applications in the table. 

## Before Changes
relying on title padding
table margin pushes it away from edge - things don't look aligned

Screenshot 2022-09-29 at 14 11 47
Screenshot 2022-09-29 at 14 12 16

Screenshot 2022-09-29 at 14 11 30

Screenshot 2022-09-29 at 14 11 12

## After changes
turn on your dev tools and explore the layout tools to see the flex grids (let me know if you need me to show you).
Screenshot 2022-09-29 at 14 16 13
Screenshot 2022-09-29 at 14 14 44
Screenshot 2022-09-29 at 14 15 15
Screenshot 2022-09-29 at 14 15 42
Screenshot 2022-09-29 at 14 15 59

Screen Shot 2022-09-29 at 14 44 50-fullpage

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -1,11 +1,11 @@
 <template>
-  <div class="rootTableContainer">
     <slot name="columnConfig" :setColumnConfig="setColumnConfig"></slot>
+    <div class="rootTableContainer">
 
       <!-- Title -->
       <div class="tableHeaderContainer">
-        <div class="title">
-            <div v-if="title">{{ title }}</div>
+        <div>
+            <p class="title" v-if="title">{{ title }}</p>
             <t-tooltip v-if="tooltip" position="top" class="tooltip" width="260px">
               <help-circle-outline-icon slot="trigger" :size="16" />
               <div class="text-sm leading-[18px] font-normal">
@@ -1232,7 +1232,9 @@ export default {
 }
 
 .rootTableContainer {
-  display: inline-block;
+  display: flex;
+  flex-direction: column;
+  gap: 15px;
   width: 100%;
   height: 100%;
   max-width: inherit;
@@ -1275,7 +1277,6 @@ export default {
 
 #tableContainer {
   display: grid;
-  margin: 5px;
   position: relative;
 }
 
@@ -1329,11 +1330,7 @@ highlighting a row on hover etc. */
   font-weight: 400;
   font-size: 20px;
   line-height: 23px;
-  display: flex;
-  align-items: center;
   font-feature-settings: "tnum" on, "lnum" on;
-  padding-top: 15px;
-  padding-bottom: 15px;
 }
 
 .tooltip{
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -1345,17 +1342,18 @@ highlighting a row on hover etc. */
...
 .tableControls {
   display: flex;
   gap: 10px;
   align-items: center;
+  justify-content: flex-end;
 }
 
-.csvExportButton{
-  min-width: 130px;
-}
```
sareg0, line 1360: This min-width resulted in extra mystery space, that was pushing away from the edge of the table

### Before change:
Screenshot 2022-09-29 at 13 29 44

If we remove this and rely on the `gap` property the effect is quite nice 

### After change:
Screenshot 2022-09-29 at 13 32 41
Screenshot 2022-09-29 at 13 32 32

visualizations/frontend/other/TTable.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <div class="rootTableContainer">
     <slot name="columnConfig" :setColumnConfig="setColumnConfig"></slot>
```
sareg0, line 2: Let me know if this will cause an bugs to put the slot at the top of the template. The slot was being considered an element and although and empty div, it was creating a gap because it was being rendered in the DOM as `<div></div>. Moving it to the top of the template took it out of the `rootTableContainer` flex container to avoid the problem.

visualizations/frontend/other/TTable.vue
msorber: The min-width was added to fix the download CSV button ending up on 2 lines when present with the modify columns button:
image

Replacing min-width in the csvExportButton class with "white-space: nowrap" should fix it though.

visualizations/frontend/other/TTable.vue
msorber: I am flabbergasted that that works in vue 2, in vue 3 you can have multiple root elements but multiple root elements in vue 2 has always resulted in a compiler error. I wonder if something is different about slots? Just to be safe and assuming doing so doesn't mess up the styling, it would be a good idea to add a div around the whole template section so there is only one root element.

PR #158, opened 2022-09-30
add clear button to relative date picker

gau94rav:
Added clear button to date picker button when a preset is selected, just like absolute dates (not part of design).

### DIFF
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -23,7 +23,7 @@
           <calendar-blank-outline-icon class="text-[16px]" />
           {{ relativePreset.label }}
           <button @click="handleClear">
-            <close-icon />
+            <close-icon title="clear date selection" />
           </button>
           <menu-down-icon class="text-[20px]" />
         </div>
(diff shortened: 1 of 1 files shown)


visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -22,6 +22,9 @@
         <div class="flex items-center gap-1">
           <calendar-blank-outline-icon class="text-[16px]" />
           {{ relativePreset.label }}
+          <button @click="handleClear">
+            <close-icon />
+          </button>
```
sareg0, line 27:
```suggestion
          <button @click="handleClear" >
            <close-icon title="clear date selection" />
          </button>
```

You can use whatever text makes sense, but without `title="clear date selection"` the button text for a screen reader would be `close icon`
I know we used a `visually-hidden` class in another component. This might replace that implementation.

sareg0 (dismissed):

visualizations/frontend/filters/v2/TDatePicker.vue
gau94rav: @sareg0 good call, thanks.

sareg0 (approved):

PR #159, opened 2022-10-03
Key value pair format change

gau94rav:
**Problem**
Tags filter uses `value_key` format for URL which can easily break with a sql layer returns `_` inside it's key and value names (value_1_key_1).

**Solution**
Replace the hacky pair system to use stringified objects in URL for easy parsing.

### DIFF
--- a/visualizations/frontend/filters/v2/TTagsInput.vue
+++ b/visualizations/frontend/filters/v2/TTagsInput.vue
@@ -110,8 +110,8 @@ export default {
             }
           }
         } catch (error) {
-          console.log(error);
-          console.log("Failed to parse tags data");
+          console.error(error);
+          console.error("Failed to parse tags data");
         }
       }
     },
--- a/visualizations/frontend/filters/v2/TTagsInput.vue
+++ b/visualizations/frontend/filters/v2/TTagsInput.vue
@@ -1,145 +1,144 @@
 <template>
-	<div class="t-tags-input max-w-max min-w-full min-h-[43px] relative">
-		<div class="pb-2 font-bold tracking-widest" v-if="label">
-			{{ label.toUpperCase() }}
-		</div>
-		<base-multiple-value-input
-			v-if="isSingleTag"
-			v-model="selectedTags"
-			selection-only
-			:data="tags.length ? tags : tempTags"
-			:loading="loading"
-			:placeholder="placeholder"
-			@update="handleUpdate"
-			display-key="value"
-			value-key="value"
-			@opened="handleOpen"
-		/>
-		<base-tags-input
-			v-else
-			v-model="selectedTags"
-			selection-only
-			:data="tags"
-			:loading="loading"
-			:placeholder="placeholder"
-			@update="handleUpdate"
-			@opened="handleOpen"
-		/>
-	</div>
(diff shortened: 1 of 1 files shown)


visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -1,145 +1,144 @@
...
+              this.tempTags = [...tags, ...this.tempTags]; // Without defaults base-multiple-value-input cannot show selected tags.
+              this.selectedTags = [...value, ...this.selectedTags]; // base-multiple-value-input needs exact value instead of key:val pairs.
+            } else {
+              this.selectedTags = [...tags, ...this.selectedTags];
+            }
+          }
+        } catch (error) {
+          console.log(error);
+          console.log("Failed to parse tags data");
+        }
```
gau94rav, line 115: Change.

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -1,145 +1,144 @@
...
+          if (!this.firstKey) {
+            break;
+          }
+          values[this.firstKey] = [item, ...(values[this.firstKey] || [])];
+        } else {
+          values[item.key] = [item.value, ...(values[item.key] || [])];
+        }
+      }
+      this.setFilterValue("selected_items", JSON.stringify(values));
+    },
```
gau94rav, line 131: Change 2.

visualizations/frontend/filters/v2/TTagsInput.vue
msorber: Does this rise to the point of using a console.error() instead of console.log()?

msorber (approved):

visualizations/frontend/filters/v2/TTagsInput.vue
gau94rav: Yes, because it means component failed entirely. updating.

msorber (approved):
LGTM! This is an excellent approach, maybe you could suggest that we make it the default in the core at the next frontend meeting?

visualizations/frontend/filters/v2/TTagsInput.vue
```diff
@@ -1,145 +1,144 @@
...
+            let tags = value.map((v) => {
+              return {
+                key,
+                value: v,
+              };
+            });
+            if (this.isSingleTag) {
+              // base-multiple-value-input is used for single tag filter.
+              this.tempTags = [...tags, ...this.tempTags]; // Without defaults base-multiple-value-input cannot show selected tags.
+              this.selectedTags = [...value, ...this.selectedTags]; // base-multiple-value-input needs exact value instead of key:val pairs.
```
sareg0, line 107: Is `value` normally a string? If yes, would the end result be what you expect I would expect the string to be split like so: 

```
const aValue = "some_value" 
const myArray = [...aValue] 
console.log(myArray)
Array(10) [ "s", "o", "m", "e", "_", "v", "a", "l", "u", "e" ]
```

sareg0 (commented):
@gau94rav would this PR need to be merged before the one in topcoat-core?

sareg0 (approved):

PR #161, opened 2022-10-11
isRequired prop added for date picker

gau94rav:
`is-required` prop can be used to hide clear dates button, to make sure there is always a default value selected.

### DIFF
--- a/visualizations/frontend/filters/v2/TDatePicker.vue
+++ b/visualizations/frontend/filters/v2/TDatePicker.vue
@@ -22,7 +22,7 @@
         <div class="flex items-center gap-1">
           <calendar-blank-outline-icon class="text-[16px]" />
           {{ relativePreset.label }}
-          <button @click="handleClear">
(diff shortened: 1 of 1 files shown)


sareg0 (approved):

visualizations/frontend/filters/v2/TDatePicker.vue
```diff
@@ -89,6 +89,10 @@ export default {
       type: String,
       default: "",
     },
+    isRequired: {
+      type: Boolean,
+      default: false,
+    },
```
sareg0, line 95: I also thought that perhaps this could be derived from `default-value` being given. I presume the default-value is there as a sort of indication that the date picker needs a value. Perhaps if a default value is provided, we should not show the reset button. However, if you've already thought on that, you can resolve this comment and move forward

gau94rav: Yes, I was also confused between both approaches, maybe in future we can have a default value that is clearable as well.

PR #174, opened 2022-11-02
add default filter slot to filter section

gau94rav:
**What's added**
A default filter slot for filter section.

**Reason**
A default filter does not have to go through the initialisation process as it is always visible, adding any filter to this slot will remove it from dropdown's init logic.

**Usage**
```
<t-filters-section>
    <t-select slot="defaultFilter" />
    // This filter will be removed from dropdown's init logic and will be directly displayed on page.
</t-filters-section>
```

### DIFF
--- a/library/components/TFiltersSection.vue
+++ b/library/components/TFiltersSection.vue
@@ -9,6 +9,7 @@
           :is="tag"
           :ref="tag"
         >
+          <slot name="defaultFilter"></slot>
           <div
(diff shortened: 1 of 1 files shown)


egcastro (approved):

egcastro:
It would be nice if you can provide an example in the PR description of the usage of this new slot. Feel free to add it after you merge the PR if you prefer.

PR #177, opened 2022-11-04
Feat/table footer

gau94rav:
Add footer slot with exposed vairables that can be used for a outer pager control or plain text.

Snyk insights [PR](https://github.com/snyk/snyk-insights-topcoat/pull/103) that uses this change.

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -244,7 +244,7 @@
       name="footer"
       :total-count="totalRows"
       :visible-count="rowsPerPage"
-      :start-ondex="startIndex"
+      :start-index="startIndex"
       :end-index="endIndex"
     ></slot>
   </div>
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -240,6 +240,13 @@
         @setResetFunction="setPagerResetFunction"
       />
     </div>
+    <slot
+      name="footer"
+      :total-count="totalRows"
+      :visible-count="rowsPerPage"
+      :start-ondex="startIndex"
```
msorber, line 247: should this be start-index instead of start-ondex?

msorber (changes requested):

visualizations/frontend/other/TTable.vue
gau94rav: @msorber thanks for catch, corrected it.

msorber (approved):

PR #178, opened 2022-11-09
Fix/t url encoding

gau94rav:
**Problem**
new URLSearchParams cannot handle space properly, it replaces it with `+`

**Fix**
Generate URL with double encoding like registry.

Note: This will be a temporary fix.


msorber (approved):
LGTM

PR #180, opened 2022-11-15
allow trigger slot customisation

gau94rav:
Allows trigger button of TNavigationDropdown to be customisable using default slot.

### DIFF
--- a/library/components/TNavigationDropdown.vue
+++ b/library/components/TNavigationDropdown.vue
@@ -1,12 +1,14 @@
 <template>
-  <t-dropdown>
-    <div slot="handle" class="flex items-center gap-1 p-1 text-sm font-medium">
-      <span style="color: #145deb" class="pl-1">
-        <chart-timeline-variant-shimmer-icon :size="18" />
-      </span>
-      {{ page.title }}
-      <menu-down-icon :size="20" />
-    </div>
+  <t-dropdown :disable-active-class="$slots.default ? true : false">
+    <slot slot="handle">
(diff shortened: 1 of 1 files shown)


library/components/TNavigationDropdown.vue
```diff
@@ -1,12 +1,14 @@
 <template>
-  <t-dropdown>
-    <div slot="handle" class="flex items-center gap-1 p-1 text-sm font-medium">
-      <span style="color: #145deb" class="pl-1">
-        <chart-timeline-variant-shimmer-icon :size="18" />
-      </span>
-      {{ page.title }}
-      <menu-down-icon :size="20" />
-    </div>
+  <t-dropdown :disable-active-class="$slots.default ? true : false">
```
sareg0, line 2: I'm curious, is this needed?

I read in the docs the following: "The default property contains either nodes not included in a named slot or contents of v-slot:default." [link](https://v2.vuejs.org/v2/api/?#vm-slots)

Wouldn't this always be true because there are nodes inside t-dropdown at all times?

gau94rav: Actually no, it will always be true if the component has a child inside it's block, the default slot simply takes any child item that does not belong to a named slot, so leaving the component empty with self closing tags will result in this statement being false.

something like this` <t-navigation-dropdown>Inner slot item</t-navigation-dropdown>` will result in true.

sareg0 (dismissed):

sareg0 (approved):

PR #184, opened 2022-12-07
add visible item count and limit for multi and tags selectors

gau94rav:
**Problem**
Currently a filter with many selected values can go out of div bouns.

**Solution**
Added a new prop called `selectedVisibleCount` that shows 5 items by default and can be customized.

[Figma](https://www.figma.com/file/1Yt1eSMomCATi3wBUeG5tr/Patch-ProductUI?node-id=10233%3A123753&t=hwTO67gRaXSBqvUZ-4)

### DIFF
--- a/visualizations/frontend/filters/TTagsSelector.vue
+++ b/visualizations/frontend/filters/TTagsSelector.vue
@@ -15,7 +15,7 @@
         <div v-if="selected.length" class="flex flex-wrap gap-1">
           <span>:</span>
           <div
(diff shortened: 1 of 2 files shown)


msorber (approved):
LGTM!

In the future it might be worthwhile to ask Nick to come up with a way to hide filters based on available size instead of count since divs may still end up out of bounds on small screen sizes and/or with large filters if I'm reading the design correctly. But that is most definitely a future problem to tackle and this is 100% an improvement and matches the current design.

PR #190, opened 2022-12-15
Feat/add remove all filters

msorber:
### What this does

This PR converts the format of filters in the topcoat core frontend from an object in the form of 
`
{
  filterName1: filterValue1,
  filterName2: filterValue2,
...
}
`

into an array in the format of:
`
  [
    { 
       name: filterName1,
       value: filterValue1,
       defaultValue: <optional default value>, 
       skipOnReset: <default to false, but can be set to true as appropriate>
    },
  ...
  ]
`

This update allows the code to reset "all" filters while leaving things like modify columns, sorting, "Issues buy" option etc. alone even though under the hood they work like filters.

Additionally it allows default values to be set on a filter, for example issue status, so that resetting the filters does not simply remove them but rather reverts the state back to the what it was before the user interacted with the page. 

It does not change the format of the filters on the backend. There is no point sending default values or whether or not a filter should be skipped on reset to the back end since all of that logic is on the front end, instead filters are sent in the original format to the back end. 

### Additional changes

Note that the addition of `...mapGetters("page", ["getFilterState"]),` in common_mixin is so that topcoat developers don't have to manually find the filter before testing its value. 

### Notes for the reviewer

To test, this requires updates in both topcoat-public and topcoat-reports so check out the 'feat/resetable_filters' topcoat-reports branch and run a build and sync modules.

### DIFF
--- a/library/components/TCsvExport.vue
+++ b/library/components/TCsvExport.vue
@@ -41,10 +41,9 @@ export default {
       };
       let filters = this.getFiltersState() ? this.getFiltersState() : [];
 
-      let formattedFilters = {};
-      filters.forEach(
-        (filter) => (formattedFilters[filter.name] = filter.value)
-      );
+  const formattedFilters = filters.reduce((acc, filter) => {
+    return {...acc, [filter.name]: filter.value}
+  }, {});
 
       if (this.additionalFilters) {
         filters = { ...formattedFilters, ...this.additionalFilters };
--- a/library/components/TCsvExport.vue
+++ b/library/components/TCsvExport.vue
@@ -36,13 +36,18 @@ export default {
   methods: {
     download() {
       this.is_loading = true;
-      payload = {
+      const payload = {
         layer: this.tLayer,
       };
-      let filters = this.getFiltersState ? this.getFiltersState : {};
+      let filters = this.getFiltersState() ? this.getFiltersState() : [];
+
+      let formattedFilters = {};
+      filters.forEach(
+        (filter) => (formattedFilters[filter.name] = filter.value)
+      );
 
       if (this.additionalFilters) {
-        filters = { ...filters, ...this.additionalFilters };
+        filters = { ...formattedFilters, ...this.additionalFilters };
       }
 
       this.trackExport("csv");
--- a/library/components/TCsvExport.vue
+++ b/library/components/TCsvExport.vue
@@ -41,9 +41,9 @@ export default {
       };
       let filters = this.getFiltersState() ? this.getFiltersState() : [];
 
-  const formattedFilters = filters.reduce((acc, filter) => {
-    return {...acc, [filter.name]: filter.value}
-  }, {});
+      const formattedFilters = filters.reduce((acc, filter) => {
+        return { ...acc, [filter.name]: filter.value };
+      }, {});
 
       if (this.additionalFilters) {
         filters = { ...formattedFilters, ...this.additionalFilters };
--- a/library/components/TFiltersSection.vue
+++ b/library/components/TFiltersSection.vue
@@ -110,8 +110,8 @@ export default {
       // To preserve original add order for filter
       // this loop is required again.
       let assignedFilters = [];
-      for (let filterIndex in this.selectedFilters) {
-        const filterName = this.selectedFilters[filterIndex].name;
+      for (const filter of this.selectedFilters) {
+        const filterName = filter.name;
         if (assignedFilters.includes(filterName)) {
           continue;
         }
@@ -167,7 +167,8 @@ export default {
           if (
             unusedOnly &&
             this.selectedFilters.find(
-              (selectedFilter) => selectedFilter.name.toLowerCase() === param.toLowerCase()
+              (selectedFilter) =>
+                selectedFilter.name.toLowerCase() === param.toLowerCase()
             )
           ) {
             urlFilters = [];
--- a/library/components/TFiltersSection.vue
+++ b/library/components/TFiltersSection.vue
@@ -180,7 +180,7 @@ export default {
       }
       return urlFilters;
     },
-    removeFilter(urlFilters, vfname) {
+    removeFilter(urlFilters, visibleFilterName) {
       for (let filterName of urlFilters) {
         const filterIndex = this.selectedFilters.findIndex(
           (f) => f.name === filterName
@@ -191,7 +191,7 @@ export default {
         this.deleteFilter({ name: filterName });
         this.filters = this.filters.filter((sf) => sf.name !== filterName);
       }
-      this.$delete(this.visibleFilters, vfname);
+      this.$delete(this.visibleFilters, visibleFilterName);
       this.handleMenuItems();
     },
     resetAllFilters() {
--- a/library/components/TFiltersSection.vue
+++ b/library/components/TFiltersSection.vue
@@ -38,10 +38,12 @@
         </component>
         <t-filters-dropdown :items="menuItems" @opened="handleFilterOpen" />
 
-
-        <div class="flex gap-1 text-[#145DEB] cursor-pointer" @click="resetAllFilters">
-            <backup-restore-icon :size="18" />
-            Reset all filters
+        <div
+          class="flex gap-1 text-[#145DEB] cursor-pointer"
+          @click="resetAllFilters"
+        >
+          <backup-restore-icon :size="18" />
+          Reset all filters
         </div>
       </div>
     </div>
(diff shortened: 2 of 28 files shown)


visualizations/frontend/filters/TTextInput.vue
```diff
@@ -271,7 +271,7 @@ export default {
       set(value) {
         this.text_internal = value;
         // Todo: Remove this when spa is live
-        this.setFilterValue("query", this.text_internal, true);
```
msorber, line 274: The setFilterValue function used to accept 3 params, that was reduced to 2 but all of the instances that called it did not get updated. Since this update increased the number of params to 3, all of those old instances needed to be cleaned up.

visualizations/frontend/filters/v2/TMultiSelector.vue
```diff
@@ -82,12 +82,15 @@
...
           <t-loading-spinner v-if="isExpanded && loading" position="relative" />
-          <span v-else>
-            {{ checked.length < menu.length ? "Select All" : "Reset" }}
+          <span
+            v-else-if="checked.length < menu.length"
+            @click="selectUnselect"
+          >
+            Select All
           </span>
+          <span v-else @click="reset"> Reset </span>
```
sareg0, line 93: Can these be buttons?

msorber: Yes!

library/components/TCsvExport.vue
```diff
@@ -36,13 +36,18 @@ export default {
...
+      const payload = {
         layer: this.tLayer,
       };
-      let filters = this.getFiltersState ? this.getFiltersState : {};
+      let filters = this.getFiltersState() ? this.getFiltersState() : [];
+
+      let formattedFilters = {};
+      filters.forEach(
+        (filter) => (formattedFilters[filter.name] = filter.value)
+      );
```
egcastro, line 47:
```suggestion
  const formattedFilters = filters.reduce((acc, filter) => {
    return {...acc, [filter.name]: filter.value}
  }, {});
```

library/components/TFiltersSection.vue
```diff
@@ -127,14 +131,11 @@ export default {
                 this.initialisedComponents.push(tag);
               }
 
-              visibleFilters[tag] = {
-                filters: urlFilters,
-              };
+              this.$set(this.visibleFilters, tag, { filters: urlFilters });
```
egcastro, line 134: Good reactivity catch ;)

library/components/TFiltersSection.vue
```diff
@@ -106,17 +110,17 @@ export default {
       // To preserve original add order for filter
       // this loop is required again.
       let assignedFilters = [];
-      const visibleFilters = {};
-      for (let filter in this.selectedFilters) {
-        if (assignedFilters.includes(filter)) {
+      for (let filterIndex in this.selectedFilters) {
```
egcastro, line 113: selectedfilters is now an Array so we should use `for...of` instead

library/components/TFiltersSection.vue
```diff
@@ -160,8 +165,10 @@ export default {
         const param = attrs[attribute];
         if (attribute.includes("t-filter")) {
           if (
-            Object.prototype.hasOwnProperty.call(this.selectedFilters, param) &&
-            unusedOnly
+            unusedOnly &&
+            this.selectedFilters.find(
+              (sf) => sf.name.toLowerCase() === param.toLowerCase()
```
egcastro, line 170:
```suggestion
              (selectedFilter) => selectedFilter.name.toLowerCase() === param.toLowerCase()
```

library/components/TFiltersSection.vue
```diff
@@ -172,16 +179,27 @@ export default {
...
       return urlFilters;
     },
-    removeFilter(urlFilters) {
-      for (let filter of urlFilters) {
-        this.deleteFilter({ name: filter });
-        if (
-          Object.prototype.hasOwnProperty.call(this.selectedFilters, filter)
-        ) {
-          this.$delete(this.selectedFilters, filter);
+    removeFilter(urlFilters, vfname) {
```
egcastro, line 182: What does `vfname` mean?

library/components/TFiltersSection.vue
```diff
@@ -172,16 +179,27 @@ export default {
...
+      for (let filterName of urlFilters) {
+        const filterIndex = this.selectedFilters.findIndex(
+          (f) => f.name === filterName
+        );
+        if (filterIndex > -1) {
+          this.selectedFilters.splice(filterIndex, 1);
         }
+        this.deleteFilter({ name: filterName });
+        this.filters = this.filters.filter((sf) => sf.name !== filterName);
```
egcastro, line 191: This shouldn't be done or needed. `this.filters` is an inherited state from the Vuex stores and the `deleteFilter` called from the line above should take care of doing exactly what this line is trying to do.

library/components/TFiltersSection.vue
egcastro: Got it, visibleFilterName I guess... We should be either more explicit or find a better naming, WDYT?

library/components/TUrl.vue
```diff
@@ -47,11 +47,13 @@ export default {
...
     fullUrl() {
       let allUrlParams = {};
+      const allFilters = {};
+      this.filters.forEach((f) => (allFilters[f.name] = f.value));
       const {
         "context[org]": org,
         "context[group]": group,
         ...restOfFilters
-      } = this.getFiltersState;
+      } = allFilters;
```
egcastro, line 56: In order to avoid future issues, we should not access `this.filters` directly but use the page's store getter instead: `getFiltersState`.
Then we can use a `reduce` call to create this object.

library/components/TUrlCopyButton.vue
```diff
@@ -28,8 +28,11 @@ export default {
     url() {
       if (this.parentUrl) {
         this.parentUrl.searchParams.set("context[page]", this.page.url);
-        for (const [filter, value] of Object.entries(this.filters)) {
-          this.parentUrl.searchParams.set(filter, encodeURIComponent(value));
+        for (const filter of this.filters) {
+          this.parentUrl.searchParams.set(
+            filter.name,
+            encodeURIComponent(filter.value)
+          );
```
egcastro, line 35: This can be a `.map` instead. I find `Array.prototype.map` more friendly than a `for...of` but this is completely subjective 😄

visualizations/frontend/other/TSelectColumns.vue
```diff
@@ -128,10 +127,15 @@ export default {
...
       const selectColumnLabels = this.checked.map((c) => c.label);
+      const defaultColumns = this?.modifiableColumns
+        .filter((mc) => mc.displayByDefault)
+        .map((mc) => mc.displayColumn);
       this.setFilter({
         name: this.urlParamName,
         value: selectColumnLabels.join("|"),
+        defaultValue: defaultColumns.join("|"),
         persist: false,
+        skipOnReset: true,
```
egcastro, line 138: 👌

visualizations/frontend/other/TTable.vue
```diff
@@ -1352,7 +1352,7 @@ export default {
           urlparam: filterName,
         });
       }
-      this.setFilterValue(filterName, filterValue);
+      this.setFilterValue(filterName, filterValue, undefined, true);
```
egcastro, line 1355: After reviewing all the code related to `setFilterValue` I think that we are at a level of need to go from single-ordered arguments to one single object argument signature.
I do not expect this to be done on these PRs but we should create a ticket so that we do not forget about it:

In `topcoat-core/visualization_mixin`, current `setFilterValue` signature:
```js
setFilterValue(name, value, defaultValue, skipOnReset=false)
```
proposal:
```js
setFilterValue({name, value, defaultValue, skipOnReset: false})
```

egcastro (changes requested):
I left some minor comments, LGTM in general

library/components/TFiltersSection.vue
msorber: Agreed. Updated to visibleFilterName

library/components/TFiltersSection.vue
msorber: Yep, removed!

visualizations/frontend/other/TTable.vue
msorber: https://snyksec.atlassian.net/browse/DP-776

library/components/TUrlCopyButton.vue
msorber: Huh. Never would have thought of using a .map there but darned if it isn't easier to read. :joy:

library/components/TCsvExport.vue
```diff
@@ -36,13 +36,17 @@ export default {
...
-      payload = {
+      const payload = {
         layer: this.tLayer,
       };
-      let filters = this.getFiltersState ? this.getFiltersState : {};
+      let filters = this.getFiltersState() ? this.getFiltersState() : [];
+
+  const formattedFilters = filters.reduce((acc, filter) => {
+    return {...acc, [filter.name]: filter.value}
+  }, {});
```
egcastro, line 46: We are missing some spaces here, indentation won't be an issue when we move this project to a proper repo+project. Sorry about it

msorber: I can't believe I missed that, thanks!

egcastro (dismissed):

egcastro (approved):

PR #193, opened 2023-01-05
Feat/default sort table prop

msorber:
To test, check out the meg-dev branch in topcoat-reports, sync and build, and open the issues detail report. The "SCORE" and "ISSUE" columns should both be sorted descending with the descending sort icon.

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -371,17 +371,18 @@ export default {
       default: () => [],
     },
     defaultSort: {
-        type: Array,
-        default: null,
-        validator(sorts) {
-            if(sorts === null) return true;
-            sorts.forEach((sort)=>{
-                if(!sort.column || typeof sort.column !== "string") return false;
-                if(!sort.direction || !["ASC", "DESC"].includes(sort.direction)) return false;
-            })
-            return true;
-        },
-    }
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TTable.vue
```diff
@@ -628,11 +624,22 @@ export default {
               );
             }
           });
-          // remove the default sort direction of columns not sorted by the url
-          sortableColumns.forEach((sc) => delete sc.direction);
           // append remaining sortable but not sorted columns to the configuration
           sortConfig = sortConfig.concat(sortableColumns);
           sortableColumns = sortConfig;
+        } else if(this.defaultSort){
```
sareg0, line 630: I am noticing that the formatting is a bit off. There are 4 spaces of indentation instead of 2 in some of the files, and here, the formatter isn't catching to add a space at the correct spot. 


```suggestion
        } else if (this.defaultSort) {
```

Your formatting extensions may need a look at. I can help if you'd like to take a look together.

sareg0 (dismissed):

visualizations/frontend/other/TTable.vue
msorber: Fixed!

sareg0 (approved):

PR #196, opened 2023-01-16
Feat/add general search to TTagsSelector

msorber:
Goal: Allow users to search for text in related fields to a filter. For example, as user may not remember "CWE-502"  but does remember "Deserialization of Untrusted Data", so even though "Deserialization of Untrusted Data" is not shown to the user as part of the filter dropdown, allow users to search for it and it should match "CWE-502"  

To test:

- check out the meg-dev branch in the builder, then run a sync and build.
- Open issues details and select the cve filter
- In the search bar of the filter, verify that filtering the CVEs by their display values works as expected
- Verify that 'SQL Injection' returns all the CVEs, but that ''SQL Injection**s**' return none
- Verify that 'Use of Hard-coded Credentials' returns no CVEs
- Remove ' :searchFields="['TITLE']" ' from the issue details page.
- Verify that 'Use of Hard-coded Credentials' returns all the CVEs

[Jira ticket](https://snyksec.atlassian.net/browse/DP-355)

[Pitch](https://www.notion.so/snyk/Enable-CVE-CWE-filter-to-include-the-description-for-easier-searching-c78f93373641466481c47fe6a471711f)

Next Steps:
Update the layers for the CWE and CVE filter layers in topcoat-reports to include title, description, etc. fields. By default, the TTagsSelector will search those fields for the user-supplied search term so no further action in the vue files will be necessary unless a field is added to the layers that should not be searchable (e.g. id), at which point all fields that should be searched will need to be added to the "searchFields" prop.

### DIFF
--- a/visualizations/frontend/filters/TTagsSelector.vue
+++ b/visualizations/frontend/filters/TTagsSelector.vue
@@ -184,6 +184,10 @@ export default {
       type: Number,
       default: 5,
     },
+    searchFields: {
+      type: Array,
+      default: null,
+    },
   },
   data: () => ({
     selected: [],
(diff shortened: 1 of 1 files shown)


visualizations/frontend/filters/TTagsSelector.vue
```diff
@@ -321,6 +313,29 @@ export default {
...
+              ?.toString()
+              .toLowerCase()
+              .includes(searchTerm.toLowerCase())
+          ) {
+            return true;
+          }
+        }
+      }
+      return false;
+    },
```
egcastro, line 338:
```suggestion
    searchInObject(searchTerm, row) {
      const COLUMN_VALUE = "value";
      const lowercaseSearchFields = this.searchFields
        ? this.searchFields.map((field) => field.toLowerCase())
        : [];
        
      for (const column of Object.keys(row)) {
        if (
          !this.searchFields ||
          lowercaseSearchFields.includes(column.toLowerCase()) ||
          column.toLowerCase() === COLUMN_VALUE
        ) {
          const data = row[column];
          
          if (
            data
              ?.value
              ?.toString()
              .toLowerCase()
              .includes(searchTerm.toLowerCase())
          ) {
            return true;
          }
        }
      }
      return false;
    },
```

egcastro: I have not tested the suggestion locally, it just a semantic change not logic change involved.

egcastro (approved):

PR #198, opened 2023-01-31
Chore/add display value option to TTagsSelector

msorber:
Goal: Give users the ability to include extra data in the filter dropdowns like CWE/CVE ID and title. After conversation with Ezra, the best way to achieve that in a flexible way is in the layer. So the TTagsSelector component will check for a DisplayValue column in the layer and show that instead of the key or value if it is present. Additionally, this approach will prevent having to update all of the TTagsSelectors where the additional functionality is not needed.

To test:

- checkout the meg-dev branch of topcoat-reports
- run a build and sync modules
- open Issues Detail page 
- add the CWE filter and  verify that the cwe IDs in the filter dropdown are all followed by " SQL Injection"
- add the CVE filter and  verify that the cve IDs in the filter dropdown are all followed by " SQL Injection"

General thoughts on this PR: It has been my experience that the less the data is manipulated in a component the fewer hacky things you have to do like manipulate date -> show it -> use that manipulated data to go find the original data to show something else. However attempting to refactor the TTagsSelector while preserving the url filters was not actually simplifying much and was therefor not worth the effort to continue. If y'all disagree I'm happy to continue with that approach instead.

[Pitch](https://www.notion.so/snyk/Enable-CVE-CWE-filter-to-include-the-description-for-easier-searching-c78f93373641466481c47fe6a471711f)


justshiv (approved):
Demoed to show manual testing and identify how this works. :shipit:

justshiv (dismissed):

visualizations/frontend/filters/TTagsSelector.vue
```diff
@@ -339,6 +341,21 @@ export default {
...
+        } else {
+          const k = r[this.tKeyColumn];
+          return k.value === key;
+        }
+      });
+      if (row?.DISPLAYVALUE) {
+        return row.DISPLAYVALUE.value;
+      }
+      return value ? value : key;
+    },
```
egcastro, line 358: At first, it was hard for me to read and understand this code, I was thinking how can we improve it, and I came up with something like this:

```suggestion
    displayValue(value, defaultValue = "") {
      if (!value) {
        return "";
      }

      const row = this.rows.find((row) => {
          const columnValue = row[this.tValueColumn];

          return columnValue.value === value;
        });

      return row?.DISPLAYVALUE?.value || defaultValue;
    },
    
    displayKey(key, defaultKey = "") {
      if (!key) {
        return "";
      }

      const row = this.rows.find((row) => {
        const columnKey = row[this.tKeyColumn];
      
        return columnKey.value === key;
      });

      return row?.DISPLAYKEY?.value || defaultKey;
    },
```

It makes the API clearer IMHO, instead of having a `value` or `key` signature, we can have dedicated methods for each. It's more verbose but it increase the simplicity.

Please, let me know what are your thoughts about it.

egcastro (dismissed):

msorber:
https://user-images.githubusercontent.com/78518576/217602586-66f7eb7d-0719-425d-ac6f-5bab5a08be55.mov

egcastro (approved):

PR #201, opened 2023-02-14
Added debounce to TTagsSelector

msorber:
Goal: Avoid race condition when fetching data where the filters have been updated but a previous result set is returned after the result for the latest severity filter change so data that does not match the selected filters is displayed to users.

To test:

- update the revision in config.yml to 'chore/add_debounce_to_TTagsSelector'
- sync and build modules
- open Issues Detail and add the severity filter
- rapidly select and deselect options and verify that the data is only fetched when there is a pause between updating the selections (the loading circles will appear when data is being fetched) and that the data stays in sync with the filters.

### DIFF
--- a/visualizations/frontend/filters/TTagsSelector.vue
+++ b/visualizations/frontend/filters/TTagsSelector.vue
@@ -285,7 +285,7 @@ export default {
         ? JSON.stringify(urlObject)
         : "";
       return this.setFilterValue("selected_items", urlValue);
-    }, 750),
+    }, 250),
     open() {
       if (this.loading) {
         return false;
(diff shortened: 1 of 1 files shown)


gau94rav (approved):
Looks good, though is 750ms debounce required? Maybe a shorter 250 should also work but anyways this shouldn't stop from merging.

visualizations/frontend/filters/TTagsSelector.vue
```diff
@@ -285,10 +285,7 @@ export default {
         ? JSON.stringify(urlObject)
         : "";
       return this.setFilterValue("selected_items", urlValue);
-    },
-    removeFilter() {
-      this.unsetFilterValue("selected_items");
-    },
+    }, 750),
```
egcastro, line 288: Are we 100% sure about this magic number? Could we test with a smaller value to see if the result is the same?

msorber: Man I need to not forget about PRs. I tested with 500 and couldn't get it to trigger but that could be me being uncoordinated. 🤷‍♀️  Happy to reduce it.

egcastro (approved):

PR #202, opened 2023-02-15
iframe fix for navigation

gau94rav:
**Problems**
- When opening a new tab from navigation dropdown inside iframe, the main URL gets used which results in redirected page to load outside of iframe.

**Fixes**

- [x] Use parent context's URL when inside iframe instead of default.
- [x] Override default click events for anchor tag to manually handle redirection.

### DIFF
--- a/library/components/TNavigationDropdown.vue
+++ b/library/components/TNavigationDropdown.vue
@@ -54,10 +54,14 @@
           class="flex justify-between mb-3 text-sm cursor-pointer hover:text-[#1c1c21]"
           :class="navSelected(page.url, 'text-[#1c1c21]', 'text-[#555463]')"
         >
-          <a :href="page.url" class="flex items-center justify-between w-full">
+          <a
+            :href="page.url"
+            class="flex items-center justify-between w-full"
+            @click.prevent="redirectToPage($event, page)"
+          >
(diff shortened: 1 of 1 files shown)


library/components/TNavigationDropdown.vue
```diff
@@ -100,11 +100,36 @@ export default {
     selected() {
       return window.location.pathname.substring(1);
     },
+    pages() {
+      const parentContextUrl = this.whoami?.context?.parent;
+      const pages = [];
+      for (let page of this.project.nav) {
+        page.main_url = page.url; // Common parameter for 'active page' check.
```
egcastro, line 107: Why do we need to create this `main_url` attribute if it's equal to `page.url`?

gau94rav: main_url actually just stores the main path of the page (issues-summary or issues-detail) while the url can get modified if inside iframe, see line 110.

gau94rav: Basically the logic is, if inside iframe then modify the default url to also use paren't full url but if anchor tag is just clicked instead of using context menu to open as new tab then `redirectToPage` function is used.

egcastro (approved):

PR #203, opened 2023-02-16
Fix priority score reset icon

icoconutgirl:
**Bug**
Close (X) icon inside priority score filter was doing nothing. 
 
**Solution**
Now this icon resets the filter to it's initial state.


gau94rav (approved):
Lgtm

PR #204, opened 2023-02-16
Feat filter search

icoconutgirl:
**Issue**
Users deleting the previous text present in search input and then searching for another filter.

**Solution**
Added autofocus and pre text selection for search input.

### DIFF
--- a/visualizations/frontend/filters/TRangeSelector.vue
+++ b/visualizations/frontend/filters/TRangeSelector.vue
@@ -15,7 +15,11 @@
           class="flex items-center gap-1 font-normal"
         >
           : {{ barMinValue }}-{{ barMaxValue }}
-          <close-icon :size="12" @click="removeFilter" />
+          <reload-icon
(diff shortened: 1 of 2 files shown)


visualizations/frontend/filters/TRangeSelector.vue
```diff
@@ -15,7 +15,11 @@
           class="flex items-center gap-1 font-normal"
         >
           : {{ barMinValue }}-{{ barMaxValue }}
-          <close-icon :size="12" @click="removeFilter" />
+          <reload-icon
```
gau94rav, line 18: These changes were already pushed in previous [PR](https://github.com/topcoat-data/topcoat-public/commit/eb46c91648a3d95e9a8ae8a0abb33f971adfe33f)

gau94rav: Only TFiltersDropdown changes required.

PR #205, opened 2023-02-16
Add autofocus and text selection

icoconutgirl:
**Issue**  
Users deleting the previous text present in search input and then searching for another filter.

**Solution**  
Added autofocus and pre text selection for search input.


gau94rav (approved):
Working great.

PR #209, opened 2023-02-23
Chore/add way to apply css classes to the table based on column, row, row index, and/or column index

msorber:
### What this does

Mehak needs a way to style the first row in table, this addresses that need in a general way. 

### Notes for the reviewer

To test:

- update revision in config.yml to "chore/add_first_and_last_row_css_classes_to_table"
- build and sync
- add ":cellCssFunction="(column, columnIndex, row, rowIndex) => { if (rowIndex === 0) return 'font-medium ' }" to a table
- verify that the first row is bold

### Missed anything?

### More information

- [Slack thread](https://snyk.slack.com/archives/)
- [Jira ticket](https://snyksec.atlassian.net/browse/TEAM-0000)

### Screenshots / GIFs

_Visuals that may help the reviewer. Please add screenshots, including URL if applicable, for any front end change. GIFs are most welcome!_

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -198,7 +198,7 @@
               :key="column.property"
               :ref="'rowCell_' + gindex + '_' + rindex + '_' + cindex"
               class="border-b border-[#D3D3D9] align-top py-[12px]"
-              :class="generateCellClasses({ column, cindex, rindex })"
+              :class="generateCellClasses({ column, cindex, row, rindex })"
             >
               <slot
                 :name="column.property"
@@ -1189,7 +1189,7 @@ export default {
     toggleRow(row) {
       row.detailRowOpen ? this.collapseRow(row) : this.expandRow(row);
     },
-    generateCellClasses({ column, cindex, rindex }) {
+    generateCellClasses({ column, cindex, row, rindex }) {
       let classes = "row ";
       classes += " cellPadding ";
       classes += _.camelCase(column.property);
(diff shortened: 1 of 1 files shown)


egcastro:
Is there anything preventing this to be done only using CSS?
Example:
```css
.row:first-child {
  font-weight: 500; //Or the CSS property needed.
}
```

egcastro:
It will look similar to:
image

msorber:
Where would that css live? It can't go in the table component itself without being applied to _every_ table and styles aren't supported in pages, instead the "style" is just amended to the page as text:
Screen Shot 2023-02-23 at 11 51 14 AM

At some point Topcoat should be updated to support styles in pages, but not without also supporting scoped styles + the deep selector or else the css will get really gnarly very quickly which is a whole different can of worms.

dbroudy:
FYI, there's support for the style tag in the topcoat component and visualization parsers, and unit tests for it:
https://github.com/snyk/topcoat-core/blob/main/tests/builder/parser/test_visualization_parser.py#L236-L244

Is there something different there than what's required here? Or is there some kind of bug? Might be an easy fix

msorber:
So adding styles to the page is not an easy fix, new pitch for adding that is [here](https://www.notion.so/snyk/Add-support-for-styles-on-pages-e5332d30cc2044fc8a81399b445fcd63). But Mehak will still need a workaround until that is completed and I think that this will be useful in general anyway for things like highlighting cells that are out of a range etc.

egcastro (approved):

PR #215, opened 2023-03-06
Fix/no data without paging

msorber:
### What this does

The table is not displaying data when paging was disabled, this fixes that issue. It isn't the best approach, but it will be a non issue when we get to [this pitch](https://www.notion.so/snyk/026dbc50a0464ccb94e389bab033235a?v=6bbfc939d0e143a781412b203e550216&p=4d632e47f425409a9d4c0b917118faad&pm=s). 

### Notes for the reviewer

- Check out the  "feat/owasp-top-10" branch of topcoat-reports
- On the owasp_top_10_report page, update: 
`:canPageServer="true"`
to
`
:canPageServer="false"
:canPage="false"
`

- In config.yml update the revision from main to `fix/no_data_without_paging` and run sync and build.
- verify that the table shows 11 lines with no paging.

### Missed anything?

### More information

- [Jira ticket](https://snyksec.atlassian.net/jira/software/c/projects/DP/issues/DP-750?jql=project%20%3D%20%22DP%22%20AND%20assignee%20%3D%20%226222996a6a4c4c0070af4496%22%20ORDER%20BY%20created%20DESC)

### Screenshots / GIFs

Screen Shot 2023-03-06 at 5 14 23 PM

### DIFF
--- a/visualizations/frontend/other/TTable.vue
+++ b/visualizations/frontend/other/TTable.vue
@@ -757,7 +757,6 @@ export default {
       if (!validProps) return;
       if (!this.rows) return;
 
-      this.setInternalColumns();
       this.setupInternalRows();
 
       // Handle radio buttons setup
@@ -821,6 +820,8 @@ export default {
           detailRowOpen: !this.canCollapseDetailRows,
         };
       });
+
+      this.setInternalColumns();
       this.setDisplayRows();
       this.showSpinner = false;
     },
@@ -1232,7 +1233,9 @@ export default {
     },
     updateEndIndex(newEndIndex) {
       this.endIndex = newEndIndex;
-      if (!this.canPageServer) this.setupInternalRows();
+      if (!this.canPageServer) {
+        this.setupInternalRows();
+      }
     },
     fetchTotalRows() {
       const payload = this.createRequestPayload();


sareg0 (approved):

visualizations/frontend/other/TTable.vue
```diff
@@ -757,7 +757,6 @@ export default {
       if (!validProps) return;
       if (!this.rows) return;
 
-      this.setInternalColumns();
```
egcastro, line 760: What's the reason we are moving this call `this.setInternalColumns()` inside the `setupInternalRows` method? Is there something that we are doing on the `setupInternalRows` that affects the execution of `setInternalColumns`?
If that the case, we should try to avoid it, if we can't, can we move swap the line orders instead of moving the call to inside the other method (so that we avoid even more coupling).

```javascript
this.setupInternalRows();
this.setInternalColumns();
```

msorber: More context, this is 100% a hack to unblock Kennan. The setInternalColumns call belongs exactly where it _was_ in init, hence why I didn't do a bigger refactor- this will either be reverted or, more likely, completely refactored when table is split up into a boring non-topcoat base table + layer handling component. Unfortunately onVisualizationInit is now being called before the row data is available (should probably be a bug ticket now that I think about it). Probably caused by the addition of a debounce in the created function in visualization_mixin in topcoat core. Additionally adding a watcher on rows is not working either- hence the hack here. It's all bad, this is just the least bad approach I can think of without having to dump a bunch more time into a solution. 🤷‍♀️

egcastro (approved):

PR #229, opened 2023-04-28
tabs component

gau94rav:
**What's new**
A new tab component for pages.

**Why use this instead of Snyk's**
We need a tab component that can support visualizations inside it's tab buttons along with label, snyk does not support that.

**Snytax**
```
<t-tabs
    t-layer="tab_layer"
    t-filter:tab="my_tab_url_parameter"
    :tabs="[
        { label: "Tab one", slot: "section_1" }, 
        { label: "Tab two", slot: "section_2" }, 
    ]">
       
        // By default tab buttons populate themselves, but to make any changes,
        // their slot name can be used which prefixed with `tab_` so section_1 becomes `tab_section_one` 
        // Tabs
       <div slot="tab_section_one" slot-scope="{tab}">
           {{ tab.label }}
       </div>

        // Sections
        <div slot="section_1">
            Tab one section 
        </div>

        <div slot="section_2">
            Tab two section 
        </div>

</t-tabs>
```

Note: `t-layer` file can just be a blank SQL file, we just need it to bind URL param using t-filter:tab attribute.

[Figma](https://www.figma.com/file/SqlXbIgBD9mpYw6JG15aQc/Account-Analytics?node-id=2560-90786&t=2WQhCzXIoBIXelB8-0)

By default this component does not have much of a design, but is entirely customizable using slots, figma is just used as a sample for functionality.

### DIFF
--- a/library/components/TTabs.vue
+++ b/library/components/TTabs.vue
@@ -1,0 +1,55 @@
+<template>
+  <div class="flex w-full" :class="!isVertical && 'flex-col'">
+    <div
+      class="flex items-center gap-2 bg-[#F9F8FA] px-4"
+      :class="isVertical && 'flex-col'"
+    >
+      <div
+        v-for="tab in tabs"
+        :key="tab"
+        class="cursor-pointer border-t-4"
+        :class="
+          activeTab === tab.slot
+            ? 'bg-white border-[#145DEB]'
+            : 'border-transparent'
+        "
+        @click="switchTab(tab)"
+      >
+        <slot :name="['tab_' + tab.slot]" :tab="tab">
+          {{ tab.label }}
+        </slot>
+      </div>
+    </div>
+    <slot v-if="activeTab" :name="activeTab"></slot>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    tabs: {
+      type: Array,
+      default: () => [],
+    },
+    isVertical: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data: () => ({
+    activeTab: null,
+    is_filter: true,
+  }),
+  methods: {
+    switchTab(tab) {
+      this.activeTab = tab.slot;
+      this.setFilterValue("tab", tab.slot);
+    },
+    onVisualizationInit() {
+      let value = this.getFilterValue("tab");
+      value = value && value !== "undefined" ? value : this.tabs[0].slot;
+      this.activeTab = value;
+    },
+  },
+};
+</script>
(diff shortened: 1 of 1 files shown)


sareg0 (commented):
Hi Gaurav, 

Can I trouble you for some screenshots of this, and or a guide on how I could test this out?
many thanks

gau94rav:
Hi @sareg0 , thanks for checking this out, for demo you can actually switch to `feat/enterprise_reporting` branch inside [Analytics page file](https://snyk-insights.topcoatdata.app/reporting/a/develop/pages/etc_analytics%7Chtml) (etc_analytics.html)

Also adding a screenshot.

gau94rav:
The tab component is made without much design to make sure it stays customizable using slots.

Default look
Screenshot 2023-05-02 at 1 52 49 PM


With customization from slots
Screenshot 2023-05-02 at 1 57 06 PM

sareg0 (dismissed):
I am happy to approve this as a first iteration. However, I have some feedback for your future iterations. I am unsure what you have already considered:

1. non-blocking: The Tab interface has no keyboard navigation possibilities. We should read up about it and improve this: https://dequeuniversity.com/library/aria/tabpanel. Given how complex the content of the tab header could be, we may need to talk with the Designers.
2. suggestion: There is no tab selected when the page is first loaded. I think you should default to the first tab as being 'active'
3. required: please add this to the documentation with explanations of the props https://docs.google.com/document/d/1pwyyc1OtS6jUoOb3xNkFRFxPgnqJQFtmmptmtQ3D358/edit#
At the very least, comment the props so that it is clear what the use case is. For example , what is the purposed of `isVertical`?
4. future consideration; add to the URL params the tab that is currently selected, so that the report can be shared with the desired tab already open.


Please let me know if you need any clarity on any of the above feedback

gau94rav:
@sareg0 Thanks for feedback Sara
1. I like the idea of keyboard navigation, maybe in next version.
2. This works when you use it as a visualization with t-layer (to bind url param), but it should still default to first item, I can add it quickly.
3. This is something that wasn't required, it switches the look of tabs and sections to be side by side instead of top and bottom, we probably don't need it though or it needs a better name.
4. Yes it works as a visualization, we need to specify any empty layer file for it, sadly this is the only way for now, as binding URL params in regular way might interfere with filters system.

gau94rav:
@sareg0 Can you please approve one more time, I have done following changes.
1. Remove is-vertical for now as it is not required and makes no sense, maybe we can plan it better in future.
2. Allow first tab to be selected if not used with a layer file.

sareg0 (approved):

PR #235, opened 2023-05-09
Feat/t tabs

gau94rav:
Multiple tabs fixes

1. Moving TTabs.vue file to visualizations folder, was accidentally stored to library, because it stores the tab value as a filter value in URL, which is important to handle active tab.

2. Force tab contents to reload using key, required for plotly.


sareg0 (approved):

sareg0:
> * Moving TTabs.vue file to visualizations folder, was accidentally stored to library, because it stores the tab value as a filter value in URL, which is important to handle active tab.
> 
> *

Does this mean the filter value is not going to be stored in the URL anymore?

Please in future always leave directions about how to test, and/or a demonstration of the problem being fixed.

gau94rav:
> > * Moving TTabs.vue file to visualizations folder, was accidentally stored to library, because it stores the tab value as a filter value in URL, which is important to handle active tab.
> 
> Does this mean the filter value is not going to be stored in the URL anymore?
> 
> Please, in future, leave directions about how to test and/or a demonstration of the problem being fixed (screenshots, recordings) etc.

@sareg0 It will be stored in URL, that's how every visualization works, I accidentally left this file inside library last time, it was meant for visualization folder instead.

If you want to test, this is the syntax.

PR #244, opened 2023-06-12
Feat/pdf table component

gau94rav:
A lightweight table component for PDF pages.

Syntax
```
<t-pdf-table t-layer="table_layer">
  
   // Columns
   <div slot="col-[COLUMN_NAME_1]" slot-scope="{ column }">
        {{ column }}
   </div>

   // Rows
   <div slot="row-[COLUMN_NAME_1]" slot-scope="{ row }">
       {{ row[COLUMN_NAME_1].rendered }}
   </div>

</t-pdf-table>
```

**Columns**
- To modify columns from slots, exposed slots can be used, that are named after their column with a prefix of `col-`.
- Column slots also expose the column name in slot prop `column`.

**Rows**
- Rows can also be modified exactly like column but instead of `col-` prefix, `row-` prefix can be used.
- Rows also have exposed `row` object to be used in slots.

**Props**
- `hiddenColumns[]`: Takes column names inside an array to hide them from table.
    Example: `['COL_1', 'COL_2']`.

- `dynamicColumns[]`: Takes array of objects to hide and show columns depending on URL parameters.
    - url_value: The column name that is associated with URL parameters.
    - sql_value: The column name that is associated with SQL layer.

    Example: In original table this prop was called `modifiableColumns`.
    ```
    [
        { url_value: 'SCORE', sql_value: 'SCORE' },
        { url_value: 'EXPLOIT MATURITY', sql_value: 'EXPLOIT_MATURITY' },
    ]
    ```

- `customColumns[]`: Not layer custom columns, takes columns name in an array, they still support `row` object to bind any data with them and are exposed using slot name exactly like other columns.
    Example: ['CUSTOM_COL_1', 'CUSTOM_COL_2']

- `rowsLimit`: Limits the number of rows to load, shows 5 by default.

- `containerClass`: Binds classes with `<table />` html tag.
- `columnClass`: Binds classes with `<th />` html tag.
- `rowsClass`: Binds classes with `<td />` html tag.

**Slots**
- `col-[COLUMN_NAME]`: Exposes any column by it's name with a `col-` prefix.
- `row-[COLUMN_NAME]`: Exposes any column by it's name with a `row-` prefix.
- `header`: Header of table, this is right before `<table />` tag starts, empty by default.
- `footer`: Footer of table, this is right after `<table />` tag ends, shows number of visible rows by default if total rows > visible rows.

Screenshot 2023-06-12 at 12 23 12 PM

### DIFF
--- a/visualizations/frontend/other/TPdfTable.vue
+++ b/visualizations/frontend/other/TPdfTable.vue
@@ -1,0 +1,134 @@
+<template>
+  <div class="relative overflow-auto">
+    <t-loading-spinner v-if="loading" position="relative" />
+    <slot name="header" :loading="loading"></slot>
+    <table class="w-full table-fixed" :class="tableClasses">
+      <thead>
+        <th v-for="column in columns" :key="column">
+          <slot :name="'col-' + column" :column="column">
+            {{ column }}
+          </slot>
+        </th>
+      </thead>
+      <tbody>
+        <tr v-for="(row, index) in rows" :key="'row-' + index">
+          <td v-for="column in columns" :key="column" class="break-all">
+            <slot :name="'row-' + column" :row="row">
+              {{ row[column] ? row[column].rendered : "" }}
+            </slot>
+          </td>
+        </tr>
+      </tbody>
+      <div v-if="!loading && !rows.length" class="text-center">No Data</div>
+    </table>
+    <slot
+      name="footer"
+      :rows-limit="rowsLimit"
+      :rows-total="rowsTotal"
+      :loading="loading"
+    >
+      <div v-if="rowsTotal > rowsLimit" class="pt-4 text-right">
+        Showing {{ rowsLimit }} of {{ rowsTotal }}
+      </div>
+    </slot>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    hiddenColumns: {
+      type: Array,
+      default: () => [],
+    },
+    dynamicColumns: {
+      type: Array,
+      default: () => [],
+    },
+    customColumns: {
+      type: Array,
+      default: () => [],
+    },
+    rowsLimit: {
+      type: Number,
+      default: 5,
+    },
+    tableClasses: {
+      type: String,
+      default: "",
+    },
+  },
+  data: () => ({
+    is_filter: true,
+    rowsTotal: 0,
+  }),
+  created() {
+    this.fetchOnMounted = false;
+  },
+  computed: {
+    columns() {
+      const columns = this.rows.length ? Object.keys(this.rows[0]) : [];
+      const filteredInternalColumns = columns.filter(
+        (col) => !this.hiddenColumns.includes(col)
+      );
+      return [...filteredInternalColumns, ...this.customColumns];
+    },
+    visibleLayerColumns() {
+      let columns = [];
+      for (let column of this.dynamicColumns) {
+        if (this.filteredColumns.includes(column.url_value)) {
+          columns.push(column.sql_value);
+        }
+      }
+      return columns;
+    },
+    filteredColumns() {
+      const layerColumnsFilterName = `${this.layer}_cols`;
+      const urlParam =
+        this.filters.filter((f) =>
+          f.name.includes(layerColumnsFilterName)
+        )[0] || {};
+      return urlParam.value ? urlParam.value.split("|") : [];
+    },
+  },
+  mounted() {
+    this.prepareFilters();
+    this.fetchLayerData();
+    this.fetchRowsCount();
+  },
+  methods: {
+    prepareFilters() {
+      this.setLayerFilter("limit", this.rowsLimit);
+      this.setLayerFilter("offset", 0); // Always showing first page on PDF, pagination not required there.
+      this.setLayerFilter("orderBy", this.getFilterState(`${this.layer}_sort`));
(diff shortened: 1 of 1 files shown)


visualizations/frontend/other/TPdfTable.vue
```diff
@@ -0,0 +1,149 @@
...
+      default: "",
+    },
+    columnsClass: {
+      type: String,
+      default: "",
+    },
+    rowsClass: {
+      type: String,
+      default: "",
+    },
```
sareg0, line 74: Can you tell me why you need these?
Would it also be acceptable to enforce a certain table style?

sareg0:
I am still seeing some rows being cut in half (in the PDF you linked), is that acceptable/expected?

Screenshot 2023-06-13 at 09 41 18

visualizations/frontend/other/TPdfTable.vue
```diff
@@ -0,0 +1,149 @@
...
+export default {
+  props: {
+    hiddenColumns: {
+      type: Array,
+      default: () => [],
+    },
+    dynamicColumns: {
+      type: Array,
+      default: () => [],
+    },
```
sareg0, line 54: Do we even need a concept of dynamicColumns in the PDF, since you cannot actually modify columns in a PDF? Can we simply use the URL params and pass them as the columns to display in the PDF?

visualizations/frontend/other/TPdfTable.vue
```diff
@@ -0,0 +1,149 @@
...
+        target: this.tag_unique,
+        layer: this.layer,
+      };
+    },
+    setLayerFilter(filterName, filterValue) {
+      const table = this.$store.state.layers.components[this.tag_unique];
+      if (!table.filters) {
+        table.filters = {};
+      }
+      window.Vue.set(table.filters, filterName, filterValue);
```
egcastro, line 145: This looks like a hack. We should add a new mutation to the layers' store to store filters. Similar to these ones: https://github.com/snyk/topcoat-core/blob/main/frontend/src/store/modules/layers.js#L397-L426

egcastro: Actually, I think that something like this should work:
```js
  this.storeAttribute({
    target: this.tag_unique,
    attribute: "filters",
    value: value,
  });
```

Where value should be a copy of the current filters plus the one you want to add:
[...currentFilter, {filterName: filterValue}]
We need to check the right shape of filters.

visualizations/frontend/other/TPdfTable.vue
gau94rav: Table keeps on evolving on every page inside multiple figmas I have seen so far, so I thought it's better to just allow full customization from outside without having to change the table from inside or force a specific design by default.

gau94rav:
> I am still seeing some rows overflowing pages (in the PDF you linked), is that acceptable/expected?
> 
> Screenshot 2023-06-13 at 09 41 18

@sareg0 cannot be fully avoided and also cannot fix this by code in any way but the old table was handling this even worse it was directing the letters in half, this new table will never do that, only a very long connected word will be pushed to second page.

visualizations/frontend/other/TPdfTable.vue
gau94rav: That was my original intention but original table uses this weird syntax that I am not really sure about.

Basically the value in URL is sometimes different from value in html.
Screenshot 2023-06-13 at 5 36 55 PM

visualizations/frontend/other/TPdfTable.vue
```diff
@@ -0,0 +1,149 @@
...
+      default: () => [],
+    },
+    dynamicColumns: {
+      type: Array,
+      default: () => [],
+    },
+    customColumns: {
+      type: Array,
+      default: () => [],
+    },
```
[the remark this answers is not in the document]
gau94rav: These are the columns that are not part of SQL file but we want to show them inside page using some sort of fixed rows data, original table again uses them, so had to add them for this table as well inside `t-column` setting, issue page is using it a lot.

visualizations/frontend/other/TPdfTable.vue
gau94rav: @egcastro Yes, that looks much cleaner, pushed it, thanks for correction.

egcastro (approved):

visualizations/frontend/other/TPdfTable.vue
sareg0: We recently had an initiative to consolidate the design of the table so we don't need to maintain so many styling variations and can maintain design consistency across reports. Here are the related PRs:
* https://github.com/topcoat-data/topcoat-public/pull/240
* https://github.com/topcoat-data/topcoat-public/pull/239

While I'm not opposed to providing these props if there is a strong need, I think they pose the problem of a table the diverges with designs by accident or on purpose. In addition, you need to repeat style props at every usage point.

sareg0 (approved):

visualizations/frontend/other/TPdfTable.vue
sareg0: It's okay. We can always address a refactor if the maintenance overhead arises.

gau94rav: Got it, thanks.

PR #269, opened 2023-09-01
ajax request watcher and pdf component changes

gau94rav:
### What this does

**Problem**
PDF Download fails on long pages, I came up with 2 fixes, one allows for increased timeout for ajax request while other replaces the backend's time.sleep(1) with axios interceptor that can track ongoing requests and signal the backend's selenium to capture the page when requests stop.

Interceptor will watch the layer requests that get fired every 250ms by frontend until layer gets the response.

1. t-pdf: Component will be able to handle ajax request's timeout from a prop (timeout=300).
2. t-pdf-download-helper: Will intercept axios to watch for layer requests, as soon as they stop it will show a component in DOM that selenium in backend will be watching for.

### Notes for the reviewer

Instructions added to backend [PR](https://github.com/snyk/topcoat-core/pull/728), only for local testing.

### Missed anything?

### More information

- [Slack thread](https://snyk.slack.com/archives/)
- [Jira ticket](https://snyksec.atlassian.net/browse/TEAM-0000)

### Screenshots / GIFs

_Visuals that may help the reviewer. Please add screenshots, including URL if applicable, for any front end change. GIFs are most welcome!_

### DIFF
--- a/library/components/TPdfDownloadHelper.vue
+++ b/library/components/TPdfDownloadHelper.vue
@@ -1,0 +1,38 @@
+<template>
+  <div v-if="isVisible" id="seleniumComponent" class="fixed top-0 bottom-0" />
+</template>
+
+<script>
+export default {
+  data: () => ({
+    responses: 0,
+    isVisible: false,
+  }),
+  watch: {
+    responses: _.debounce(function (newVal) {
+      this.showSeleniumComponent();
+    }, 1000),
+  },
+  created() {
+    this.countAjaxResponses();
+  },
+  methods: {
+    countAjaxResponses() {
+      var $this = this;
+      axios.interceptors.response.use(
+        function (response) {
+          $this.responses += 1;
+          return response;
+        },
+        function (error) {
+          $this.responses += 1;
+          return Promise.reject(error);
+        }
+      );
+    },
+    showSeleniumComponent() {
+      this.isVisible = true;
+    },
+  },
+};
+</script>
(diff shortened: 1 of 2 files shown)


library/components/TPdfDownloadHelper.vue
```diff
@@ -0,0 +1,38 @@
...
+<script>
+export default {
+  data: () => ({
+    responses: 0,
+    isVisible: false,
+  }),
+  watch: {
+    responses: _.debounce(function (newVal) {
+      this.showSeleniumComponent();
+    }, 1000),
```
dbroudy, line 14: This would just wait for no ajax requests for 1 second?

dbroudy (commented):
I thought the idea was to count specific outgoing requests (layers++) and responses (layers--), and when we get to zero, show the component.

library/components/TPdfDownloadHelper.vue
gau94rav: Yes, basically ajax keeps sending requests to pdf worker for response at the interval of 500ms until it receives 200, so this component is using debounce to wait until there is no request within 1 second and it will stop and show the component for selenium.

dbroudy: this could include other non-relevant requests too, like analytics or something

gau94rav: Just checked it does not include them because we are just intercepting axios (which is used internally) but if there is another call added to axios that is not part of layers, then it would cause problems, so I will just filter out layers request only, thanks for pointing out.

dbroudy (approved):

PR #283, opened 2023-12-15
fix: disable export to csv when there is not data

egcastro:
### What this does

When there is not data, the CSV download process fails. In order to avoid this, we disable the button until there are results.

### Notes for the reviewer

_Instructions on how to run this locally, background context, what to review, best to review commit-by-commit?, questions…_

### More information

https://linear.app/snyk/issue/FP-1788/nosuchkey-displayed-when-downloading-csv-from-reports-page

### Screenshots / GIFs

Enabled:
image

Disabled:
image

### DIFF
--- a/library/components/TCsvExport.vue
+++ b/library/components/TCsvExport.vue
@@ -2,10 +2,10 @@
   <button
     class="rounded cursor-pointer border-1 border-[#C3C2CB] text-[#555463] bg-white w-max p-1"
     style="height: fit-content"
-    :disabled="is_loading"
+    :disabled="is_loading || disabled"
     @click="download"
   >
-    <div class="flex items-center gap-1">
+    <div class="flex items-center gap-1" :class="{ disabled }">
       <t-loading-spinner v-if="is_loading" position="relative" />
       <download-outline-icon v-else :size="18" />
       {{ label }}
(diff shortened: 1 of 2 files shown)


library/components/TCsvExport.vue
```diff
@@ -73,3 +78,12 @@ export default {
...
 </script>
+
+<style scoped>
+.disabled {
+  background-color: #f2f1f4;
+  border-color: transparent;
+  color: #b3b2bd;
+  cursor: default;
+}
+</style>
```
sareg0, line 89: Did you already try applying the styles to the disabled pseudo-class? I ask because there would be only one piece of data to use to style the button and put it into a disabled state.

e.g.
```
<button class="csv-download..."

csv-download:disabled {
  background-color: #f2f1f4;
  border-color: transparent;
  color: #b3b2bd;
  cursor: default;
}
```

I think it's also possible to do it via tailwaind, but I understand you may be doing this for better compatibility with reports-ui (to avoid all the arbitrary classes).


non-blocking:
I don't like the styling of the button very much when it's disabled. I wonder if not displaying the button would be better? You can go with this, and we can improve.

sareg0 (approved):
Minor, non-blocking comments.

sareg0 (approved):

